PR #50, opened 2014-12-27
Twitch.tv provider

themasch:
I added this very basic oauth2 provider for twitch.tv. 
Docs for the twitch.tv api can be found here: https://github.com/justintv/Twitch-API/blob/master/authentication.md

I might improve this provider in the future (like using `/user/:username` instead of `/user` to get the profile if user_read isn't set). But it should work good as it is now.

### DIFF
--- a/lib/providers/twitch.js
+++ b/lib/providers/twitch.js
@@ -9,6 +9,7 @@ exports = module.exports = function (options) {
         token: url + 'oauth2/token',
         scope: [ 'user_read' ],
         profile: function(credentials, params, get, callback) {
+
             // twitch requires 'OAuth' instead of 'Bearer'
             var twOptions = {
                 headers: {
--- a/lib/providers/twitch.js
+++ b/lib/providers/twitch.js
@@ -2,6 +2,7 @@ var Hoek = require('hoek');
 var url = 'https://api.twitch.tv/kraken/';
 
 exports = module.exports = function (options) {
+
     return {
         protocol: 'oauth2',
         auth: url + 'oauth2/authorize',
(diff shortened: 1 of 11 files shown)


lib/providers/twitch.js
```diff
@@ -0,0 +1,27 @@
+var Hoek = require('hoek');
```
geek: Please add to the top:
```
// Load modules
```

lib/providers/twitch.js
```diff
@@ -0,0 +1,27 @@
...
+var url = 'https://api.twitch.tv/kraken/';
```
geek: Please use the url string value in the return object below instead of a variable.

lib/providers/twitch.js
```diff
@@ -0,0 +1,27 @@
...
+
+            // twitch requires 'OAuth' instead of 'Bearer'
+            var twOptions = {
+                headers: {
+                    Authorization: 'OAuth ' + params.access_token
+                }
+            };
+            Hoek.merge(this.provider, twOptions);
+
+            get(url + 'user', {}, function(body) {
```
geek: `function (body) {`

lib/providers/twitch.js
```diff
@@ -0,0 +1,27 @@
...
+                credentials.profile = body;
```
geek: newline after function declarations

geek:
Looks good.  Please make the minor changes and remerge and its good to go!

geek:
@themasch do you think you will have time to get this PR updated?

themasch:
Meh, the merge broke stuff...

geek:
@themasch ya, you may be better off making a new PR after pulling the latest into a new branch and applying your changes.

themasch:
okay, i'll do that then. I close this one here, not needed anymore.

Issue #40, opened 2014-11-13
Forcing redirect_uri to be https


briandela:
Just wondering if it's possible to force to `redirect_uri` to be https. We're deploying hapi on Heroku and heroku does SSL termination at the edge of their network. This means we run hapi as http. Bell seems to take the protocol the site is running as and make that the protocol for the redirect_uri - this causes an issue when the provider (e.g. facebook) redirect back to our site as it comes back as http instead of https.

I couldn't see in the options a way to tell bell to use https for the redirect_uri protocol. Just wondering if there is a work around for the moment.

petehunt:
+1, every app server deployment I've worked with terminates SSL with nginx

eiriksm:
+1 from me. Had me wondering for a while. Would a PR implementing this as an option be considered?

eiriksm:
Instead of waiting for an answer to that, I went ahead and submitted the PR over at #53

ericeslinger:
This doesn't seem to work yet- I tried bell 2.1.0 with forceHttps: true, and bell threw 500 errors with "redirect_uri_mismatch" responses, which seem to be coming off of the google endpoint.

I rolled back to bell 2.0.0 and used @fampinheiro 's workaround, and it seems to work. In earlier versions (hapi 7.5), I was able to just set the location parameter on the server, and bell seemed to pick it up. However, hapi 8.0 dropped support for location, and setting uri at server creation doesn't seem to work.

I'll spend some time tomorrow actually diagnosing the problem, tonight I'm just too worn out from chasing this down- we're upgrading our beta servers from the hapi 7 to 8 version, and I thought I had everything, but https is one of those things you don't really notice on your typical dev laptop.

KrisSiegel:
I'm having the exact same issue as @ericeslinger; I upgraded to 2.1.0 but whenever forceHttps is being used google returns "redirect_uri_mismatch". If I find a cause in bell I'll report back as this is something I'm actively debugging.

KrisSiegel:
Okay yeah the internals.location method is used in several places but in the PR it's not updated in every place to use the protocol (which may be a forced https). As a debug step I simply updated internals.location to do this and it worked perfectly:
```
internals.location = function (request, protocol) {
    return "https://<myhost>/auth/google";
    var info = request.connection.info;
    protocol = protocol || info.protocol;
    var host = request.info.host || (info.host + ':' + info.port);
    return protocol + '://' + host + request.path;
};
```
[...]

PR #53, opened 2015-01-17
Add option for forcing https.

eiriksm:
As mentioned in #40, this is a PR adding the option to force https in redirect_uri.

Granted, it might not be the best solution to alter the request object like this, but this certainly works. The other option would be to pass in the settings as a parameter to the location function, not sure which one you prefer? IMO the naming of the option (force) indicates some force, and thus we are forcing the request object to be https :)

PR also adds a test for this.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -135,6 +135,10 @@ exports.v2 = function (settings) {
           return reply(Boom.internal('App was rejected'));
         }
 
+        // Alter the request to be able to force HTTPS.
+        if (settings.forceHttps) {
+            request.connection.info.protocol = 'https';
+        }
         // Sign-in Initialization
 
         if (!request.query.code) {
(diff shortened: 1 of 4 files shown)


peny:
+1

lib/oauth.js
```diff
@@ -135,6 +135,10 @@ exports.v2 = function (settings) {
           return reply(Boom.internal('App was rejected'));
         }
 
+        // Alter the request to be able to force HTTPS.
+        if (settings.forceHttps) {
+            request.connection.info.protocol = 'https';
```
geek: Instead, we should pass a new parameter to `location(request)` for `protocol`.  Otherwise we are overwriting request connection information that may not always be safe.

eiriksm:
Added a new commit that passes the protocol to the location function instead (as suggested in the diff comment. And by myself - for that matter :))

geek:
@eiriksm thanks for the help, looks good.  I'll publish soon

PR #60, opened 2015-02-21
Make forceHttps complete 

dapus:
`redirect_uri` is sent to the provider a second time. This code forces https (if option is set) in that case as well.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -176,12 +176,17 @@ exports.v2 = function (settings) {
             return reply(Boom.internal('Incorrect ' + name + ' state parameter'));
         }
 
+        var protocol = request.connection.info.protocol;
+        if (settings.forceHttps) {
+          protocol = 'https';
+        }
+
         var query = {
             client_id: settings.clientId,
             client_secret: settings.clientSecret,
             grant_type: 'authorization_code',
             code: request.query.code,
-            redirect_uri: internals.location(request)
+            redirect_uri: internals.location(request, protocol)
         };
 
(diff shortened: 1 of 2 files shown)


lib/oauth.js
```diff
@@ -176,12 +176,17 @@ exports.v2 = function (settings) {
             return reply(Boom.internal('Incorrect ' + name + ' state parameter'));
         }
 
+        var protocol = request.connection.info.protocol;
+        if (settings.forceHttps) {
+          protocol = 'https';
```
geek: we are somewhat in the minority on this, but we use 4 spaces.

test/oauth.js
```diff
@@ -509,7 +509,17 @@ describe('Bell', function () {
                     server.inject('/login', function (res) {
 
                         expect(res.headers.location).to.contain(mock.uri + '/auth?special=true&client_id=test&response_type=code&redirect_uri=https%3A%2F%2Flocalhost%3A80%2Flogin&state=');
-                        mock.stop(done);
+                        var cookie = res.headers['set-cookie'][0].split(';')[0] + ';';
+
+                        mock.server.inject(res.headers.location, function (res) {
+                            expect(res.headers.location).to.contain('https://localhost:80/login?code=1&state=');
```
geek: after function declarations we expect a newline: http://hapijs.com/styleguide

dapus:
Thanks for the feedback. This is according to the styleguide, right? :rocket:

geek:
@dapus ya, the hapijs style guide.

PR #61, opened 2015-02-25
Added protocol option to the auth_callback

KrisSiegel:
After some debugging I found forceHttps wasn't actually working and one of the internals.location() calls was missing the protocol parameter. So simply adding that in seems to have resolved the issue hence the PR.

This was debugged and referenced in #40.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -37,8 +37,12 @@ exports.v1 = function (settings) {
         if (!request.query.oauth_token) {
 
             // Obtain temporary OAuth credentials
+            var protocol = request.connection.info.protocol;
+            if (settings.forceHttps) {
+                protocol = 'https';
+            }
 
-            var oauth_callback = internals.location(request);
+            var oauth_callback = internals.location(request, protocol);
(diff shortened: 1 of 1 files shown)


KrisSiegel:
Hmm it seems I do not understand how the test code coverage stuff works in this library (I had thought it was covered with the other tests). Unfortunately I'm a bit under the gun on a current project and can't really work on this further; can someone assist in fixing the code coverage?

geek:
@KrisSiegel have you tried the latest 2.2.0 version yet?  https://github.com/hapijs/bell/pull/60

KrisSiegel:
Oh well would you look at that. I had 2.1.0 set as of yesterday morning before 2.2.0 was released, experienced the issue, forked the repo (which was 2.2.0 at that time), did some debugging, thought I fixed it but it looks like 2.2.0 actually fixed it and not me.

That's pretty funny. Okay this pull request is useless. Thanks!

geek:
@KrisSiegel thanks for the effort, happy its resolved the issue!!!

Issue #25, opened 2014-09-22
Test Cleanup


arb:
Clean up test to be in line with new standards.

ldesplat:
@arb I don't know what this referenced and it's been a long time. Is this still relevant?

gergoerdosi:
@ldesplat See here: https://github.com/hapijs/discuss/issues/24

PR #65, opened 2015-03-29
Add ArcGISOnline provider

robertd:
I'll revisit tests once #25 has been addressed.
Example works once you put in your client id & secret.

Cheers


lib/index.js
```diff
@@ -13,7 +13,6 @@ var internals = {};
 
 
 exports.register = function (server, options, next) {
-
```
geek: we use newlines after each new scope

robertd: I will revert that one

robertd:
Squashed and force-pushed

geek:
This looks good, please add this to the list in the README: https://github.com/hapijs/bell#options

geek:
@robertd can't be merged, please update

PR #73, opened 2015-04-14
Allow environment variables to configure options

### LINKED ISSUE #72: Allow environment variables to configure scheme options
Currently, it is not possible to configure some of the options in this plugin using environment variables, since some of the options are assumed to be boolean/number types, while environment variables are strings.

I would like to request that this be made possible by using Joi's nifty validation() function which will accept a boolean/number string and return a transformed object with the appropriate types.

Pull request incoming.

hofan41:
Closes #72

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -5,8 +5,8 @@ var Code = require('code');
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Lab = require('lab');
-var Mock = require('./mock');
 var Sinon = require('sinon');
+var Mock = require('./mock');
 var OAuth = require('../lib/oauth');
 
 // Declare internals
@@ -24,7 +24,7 @@ var expect = Code.expect;
 
 describe('Bell', function () {
 
-    it('supports string representations of boolean and number strategy options', function(done) {
+    it('supports string representations of boolean and number strategy options', function (done) {
 
         var mock = new Mock.V1();
         mock.start(function (provider) {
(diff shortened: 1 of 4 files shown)


hofan41:
Issue #72

geek:
This looks good.  Please add a couple of tests for making the environment variable be passed in.  Also, can you add details to the documentation on using environment variables with bell?

hofan41:
Hi geek,
I am at an impasse. 

How would you recommend I approach the tests? I can either copy an existing test that tests functionality that relies on a boolean input option, or I can mock/spy within the scheme function and confirm that the cookie settings are correctly interpreted. 

There are pros and cons to both: copying an existing test is easy, but not easy to scale and becomes a maintenance problem. mock/spy is not readily available because the server object passed to the scheme function is not the same server object that is supplied by the test (hapi uses a server._clone).

hofan41:
Committed the mock/spy test.

test/index.js
```diff
@@ -6,7 +6,8 @@ var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Lab = require('lab');
 var Mock = require('./mock');
-
+var Sinon = require('sinon');
```
geek: Please move this above the `Mock` require

test/index.js
```diff
@@ -23,6 +24,48 @@ var expect = Code.expect;
 
 describe('Bell', function () {
 
+    it('supports string representations of boolean and number strategy options', function(done) {
```
geek: `function (done) {`

test/index.js
```diff
@@ -23,6 +24,48 @@ var expect = Code.expect;
...
+                    ttl: '1234567890',
+                    forceHttps: 'true',
+                    provider: provider,
+                    config: {
+                        testNoBoolean: 'false',
+                        testNoNumber: '0987654321'
+                    }
+                });
+
+                expect( spy.calledOnce ).to.be.true();
```
geek: `expect(spy.calledOnce).to.be.true();`

no need for extra spaces in ()

geek:
Looks good, just a few minor style items

PR #77, opened 2015-04-16
Support for Phabricator

### LINKED ISSUE #75: Add phabricator third party provider
Phabricator is a useful PM tool that we have been using that also ships with an oauth 2.0 api. 

http://www.phabricator.com

hofan41:
Added support for third party installations of phabricator, the user must supply a URI in the config options in order for it to work.

Closes #75

### DIFF
--- a/lib/providers/phabricator.js
+++ b/lib/providers/phabricator.js
@@ -35,7 +35,7 @@ exports = module.exports = function (options) {
                     username: profile.result.userName,
                     displayName: profile.result.realName,
                     email: profile.result.primaryEmail,
-                    raw: profile.result
+                    raw: profile
                 };
 
                 return callback();
--- a/lib/providers/phabricator.js
+++ b/lib/providers/phabricator.js
@@ -1,0 +1,45 @@
+// Load modules
+var Joi = require('joi');
+var Hoek = require('hoek');
+
+// Declare internals
+
+var internals = {};
+
+internals.schema = Joi.object({
+    uri: Joi.string().uri().required()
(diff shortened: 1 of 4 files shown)


lib/providers/phabricator.js
```diff
@@ -0,0 +1,45 @@
...
+    var settings = results.value;
+
+    return {
+        protocol: 'oauth2',
+        auth: settings.uri + '/oauthserver/auth/',
+        token: settings.uri + '/oauthserver/token/',
+        scope: ['whoami'],
+        scopeSeparator: ',',
+        profile: function (credentials, params, get, callback) {
+            var query = {
```
geek: we add a newline after function declarations.  please update the PR with the newlines. Here is more detail on the styleguide: hapijs.com/styleguide

test/providers.js
```diff
@@ -1689,4 +1689,104 @@ describe('#linkedin', function () {
...
+            var mock = new Mock.V2();
+            mock.start(function (provider) {
+
+                var server = new Hapi.Server();
+                server.connection({ host: 'localhost', port: 80 });
+                server.register(Bell, function (err) {
+
+                    expect(err).to.not.exist();
+
+                    expect( function() {
```
geek: `expect(function () {`

geek: you can rewrite as

`expect(Bell.providers.phabricator).to.throw(Error);`   <-- I think

hofan41: I dig that. Updating now!

Issue #66, opened 2015-04-02
Deprecated Google and Facebook APIs


ulrik59:
Hi everyone,

I'm actually using Bell for Facebook and Google without any problem but I noticed that used API for both providers are deprecated.
Here's the documentation about Google: https://developers.google.com/accounts/docs/OAuth2LoginV1
And Facebook send this warning when I want to authenticate: 
![deprecatedfacebook](https://cloud.githubusercontent.com/assets/6003644/6968793/adb6450a-d969-11e4-9164-026109a3387f.jpg)

Is there any plan to update Facebook and Google providers code ?
Thank you

geek:
Thanks.  We definitely need to fix that if there is an updated API we should be using.

geek:
The Facebook issue is fixed in 2.4.0

geek:
Google is fixed by https://github.com/hapijs/bell/pull/108 in bell v5.0.0

PR #78, opened 2015-04-21
update facebook URLs to use facebook Graph v2.3

beeant:
might be a fix for #66 

reference:
- https://developers.facebook.com/docs/apps/upgrading
- https://github.com/jaredhanson/passport-facebook/commit/c5a3fa65685eae74de2b68564da7f3c6c440111c

### DIFF
--- a/lib/providers/facebook.js
+++ b/lib/providers/facebook.js
@@ -12,7 +12,7 @@ exports = module.exports = function (options) {
 
     return {
         protocol: 'oauth2',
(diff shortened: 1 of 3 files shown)


geek:
@beeant the tests are failing :/  Mind taking a look at those as they relate to this change.

beeant:
@geek I forgot to update the URLs in `test/oauth.js`. I just changed them. It passed the test now. Thank you.

PR #85, opened 2015-05-04
add location setting similar to old hapi server location

brand7n:
The `forceHttps` option is not sufficient to cover every configuration.  For example the port visible to the user's browser may be different than the port inferred by the Hapi server.  In fact, the host may also be different if there is a load balancer involved.  This PR attempts to restore the flexibility available when bell referenced the Hapi server location setting.


jplock:
+1

suicidejack:
+1

erickrdch:
+1

geek:
@brand7n well done, this looks great :)

brand7n:
@geek awesome...thanks!!

PR #87, opened 2015-05-10
Pass `forceHttps` and `location` to v1 client

salzhrani:
Pass `forceHttps` and `location` to v1 client

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -507,7 +507,6 @@ internals.location = function (request, protocol, location) {
         return location + request.path;
     }
     var info = request.connection.info;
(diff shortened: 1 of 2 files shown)


salzhrani:
coverage failing on line 510 ... not sure how to fix

geek:
You should be fine to delete line 510 since you are always passing in the protocol from `info.protocol`

If you look at the html report for coverage it should show what is needed.

salzhrani:
Thanks for the feedback,It is green now. Hope to see this land soon.

geek:
@salzhrani this is now published with v2.7.1.  Thanks

PR #91, opened 2015-05-26
OAuth client resource() endpoint

### LINKED ISSUE #90: Expose OAuth 1.0 client protected resource endpoing
Current client provides a `get()` method that automatically parses the response. I need a way to get back the response stream untouched as well as have access to methods other than `GET`.

hueniverse:
Closes #90 

This is a bit convoluted but required to keep it backwards compatible with minimal code repeats. In future breaking versions it will be worth getting rid of the `request()` and `get()` methods and just keep the new `resource()` one.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -304,34 +304,45 @@ internals.Client.prototype.temporary = function (oauth_callback, callback) {
 };
 
 
-internals.Client.prototype.token = function (oauth_token, oauth_verifier, tokenSecret, callback) {
+internals.Client.prototype.token = function (oauthToken, oauthVerifier, tokenSecret, callback) {
 
     // Token Credentials (2.3)
 
     var oauth = {
-        oauth_token: oauth_token,
-        oauth_verifier: oauth_verifier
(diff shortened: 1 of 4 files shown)


lib/oauth.js
```diff
@@ -304,34 +304,45 @@ internals.Client.prototype.temporary = function (oauth_callback, callback) {
...
+    this._prepare(method, uri, params, oauth, options, function (method, uri, header) {
+
+        Wreck.request(method, uri, { headers: { Authorization: header } }, callback);
+    });
+};
 
-internals.Client.prototype.request = function (method, uri, params, oauth, tokenSecret, type, callback) {
 
-    var provider = this.provider;
+internals.Client.prototype._prepare = function (method, uri, params, oauth, options, next) {
```
geek: is it clearer as `callback` instead of `next` ?

geek:
Looks good, just one minor question on style.

lib/oauth.js
hueniverse: `callback` must be called on next tick, `next` does not.

PR #94, opened 2015-05-27
Rewrite OAuth 1.0 client interface

### LINKED ISSUE #92: Fix POST payload format and location
The client `resource()` method puts POST payloads in the query. Also, some providers (I'm looking at your stupid face Twitter) require the form-encoded payload to be OAuth compliant which means strings with `'!'` will not get encoded using normal encoding and break. This fixes it in a way that will work under all versions of node.

If you didn't use the `Bell.oauth.Client` API, you are not affected by this change. If you did, you need to replace calls to `client.request()` with the new `client.resource()`.

hueniverse:
Closes #92

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -38,6 +38,7 @@ exports.v1 = function (settings) {
         if (!request.query.oauth_token) {
 
             // Obtain temporary OAuth credentials
+
             if (settings.forceHttps) {
                 protocol = 'https';
             }
@@ -59,7 +60,7 @@ exports.v1 = function (settings) {
                 var query = settings.providerParams ? Hoek.clone(settings.providerParams) : {};
                 query.oauth_token = payload.oauth_token;
 
-                return reply.redirect(settings.provider.auth + '?' + Querystring.encode(query));
+                return reply.redirect(settings.provider.auth + '?' + internals.queryString(query));
             });
         }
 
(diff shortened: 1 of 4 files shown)


lib/oauth.js
```diff
@@ -354,32 +344,50 @@ internals.Client.prototype._prepare = function (method, uri, params, oauth, opti
...
+        }
+    }
 
-    this._prepare(method, uri, params, oauth, { secret: tokenSecret }, function (method, uri, header) {
+    if (options.stream) {
+        return Wreck.request(method, uri, requestOptions, callback);
+    }
 
-        Wreck[method.toLowerCase()](uri, { headers: { Authorization: header } }, function (err, res, payload) {
+    var desc = (options.desc || 'resource');
```
cjihrig: The parens aren't needed.

lib/oauth.js
```diff
@@ -532,3 +540,44 @@ internals.location = function (request, protocol, location) {
...
     return protocol + '://' + host + request.path;
 };
+
+
+// Provide own QS implementation for cross node version support
+
+internals.encodePrimitive = function (value) {
+
+    var type = typeof value;
+    if (type === 'boolean') {
```
cjihrig: I've heard that V8 can better optimize `typeof var === 'value'` patterns, so it might be preferable do multiple `typeof`s instead of caching the type.

lib/oauth.js
hueniverse: I know. I just like them for readability. It breaks the `=` from the rest of the statement.

lib/oauth.js
hueniverse: I would stay away from micro optimizations. This code should go away once this no longer supports node 0.10 because we can use the querystring encode override option in 0.12 and io.

PR #96, opened 2015-06-02
Allow null for domain

briandela:
The default value for `domain` is `null`. However the Joi validation does not accept `null` as valid. Plugins like crumb/yar allow null to be passed in for this. PR is to make it consistent to enable easily using the same variable for `domain` across multiple plugins.


geek:
This looks like a good change.  It may seem like overkill for this case, but can you add a test that passes in `null` to verify this change and prevent future regressions of this setting.

briandela:
@geek - I've updated with the test.

PR #97, opened 2015-06-03
Updating Readme examples to work with current version of hapi

ryanmurakami:
Removing port in the constructor and adding call to connection.


geek:
Nice catch!

Issue #98, opened 2015-06-05
Don't send client credentials as parameters and basic auth while retrieving token


ldesplat:
I am using WSO2 and bell. I see that in bell in https://github.com/hapijs/bell/blob/master/lib/oauth.js#L191-L203

We send both the client_id and client_secret in the query and in the Authorization header. Looking at the source code of WSO2 it says that the standard does not allow both ways to be sent at the same time. I looked over the standard. [rfc6749](https://tools.ietf.org/html/rfc6749) especially section 2.3.1 and section 5.2 invalid_client . I do not see that listed.

WSO2 does not allow that behavior.

Either way, I think this could be exposed in another option when defining the provider (credTransport: ['basic', 'query']) with both as default to be backwards compatible? Better name? Better place to put it?

Thank You.

ldesplat:
And of course I missed that the standard says:
`The client MUST NOT use more than one authentication method in each request.`

So, because of that I plan on implementing this as a breaking change. I am going to add a setting in the provider named `authMethod` which will be required when protocol is 'oauth2' and will have no default. It will have only 2 valid values `basic` or `param`. 

This unfortunately breaks all the providers but a lot of them are well documented.

PR #99, opened 2015-06-05
Ensure only 1 authentication method is used during /token access in o…

ldesplat:
…auth2

Added authMethod in provider options schema which is required when oauth2 is selected.

Now, when retrieving the token bell will either use body parameters or basic header auth but not both at the same time. See Issue #98 .

I have updated all the oauth2 providers as per their specs:

[arcgisonline](http://resources.arcgis.com/en/help/arcgis-rest-api/index.html#//02r30000009z000000)
[dropbox](https://www.dropbox.com/developers/core/docs#oauth2-methods)
[facebook](https://developers.facebook.com/docs/facebook-login/access-tokens#apptokens)
[foursquare](https://developer.foursquare.com/overview/auth)
[github](https://developer.github.com/v3/oauth/)
[google](https://developers.google.com/identity/protocols/OpenIDConnect)
[instagram](https://instagram.com/developer/authentication/)
[linkedin](https://developer.linkedin.com/docs/oauth2)
[live](https://msdn.microsoft.com/en-us/library/hh243647.aspx)
[nest](https://developer.nest.com/documentation/cloud/how-to-auth)
[phabricator](https://secure.phabricator.com/book/phabcontrib/article/using_oauthserver/)
[reddit](https://github.com/reddit/reddit/wiki/OAuth2)
[vk](https://vk.com/dev/auth_sites)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ The `server.auth.strategy()` method requires the following strategy options:
         - `'oauth'` - OAuth 1.0a
         - `'oauth2'` - OAuth 2.0
     - `temporary` - the temporary credentials (request token) endpoint (OAuth 1.0a only).
-    - `useParamsAuth` - boolean that determines if OAuth client id and client secret will be sent as parameters as opposed to an Authorization header (OAuth 2.0 only). Defaults to false.
+    - `useParamsAuth` - boolean that determines if OAuth client id and client secret will be sent as parameters as opposed to an Authorization header (OAuth 2.0 only). Defaults to `false`.
     - `auth` - the authorization endpoint URI.
     - `token` - the access token endpoint URI.
     - `version` - the OAuth version.
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ The `server.auth.strategy()` method requires the following strategy options:
         - `'oauth'` - OAuth 1.0a
         - `'oauth2'` - OAuth 2.0
     - `temporary` - the temporary credentials (request token) endpoint (OAuth 1.0a only).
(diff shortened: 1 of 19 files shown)


lib/index.js
```diff
@@ -38,6 +38,7 @@ internals.schema = Joi.object({
         protocol: Joi.string().valid('oauth', 'oauth2'),
         temporary: Joi.string().when('protocol', { is: 'oauth', then: Joi.required(), otherwise: Joi.forbidden() }),
         auth: Joi.string().required(),
+        authMethod: Joi.string().valid('basic', 'param').when('protocol', { is: 'oauth2', then: Joi.required(), otherwise: Joi.forbidden() }),
```
hueniverse: I would implement this as a boolean since there are only two ways to authenticate. I would make the default using the header because that's the proper way to do it and allow for a params override.

ldesplat:
@hueniverse Updated as per your comments. Thank you!

README.md
```diff
@@ -71,6 +71,7 @@ The `server.auth.strategy()` method requires the following strategy options:
         - `'oauth'` - OAuth 1.0a
         - `'oauth2'` - OAuth 2.0
     - `temporary` - the temporary credentials (request token) endpoint (OAuth 1.0a only).
+    - `useParamsAuth` - boolean that determines if OAuth client id and client secret will be sent as parameters as opposed to an Authorization header (OAuth 2.0 only). Defaults to false.
```
hueniverse: `false`

ldesplat:
Thanks, done.

PR #101, opened 2015-06-15
Fix clientSecret usage

qualquervalor:
The code was attempting to access the clientSecret with the wrong variable name.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -196,7 +196,7 @@ exports.v2 = function (settings) {
 
         if (settings.provider.useParamsAuth) {
             query.client_id = settings.clientId;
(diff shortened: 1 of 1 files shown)


ulrikaugustsson:
The second commit should probably not be in the same PR.

cjihrig:
@geek LGTM

hofan41:
This was a rather large bug that somehow slipped through the unit tests. Wonder why the unit tests didn't catch this.

cjihrig:
Likely because tests use mock data. PRs adding more/better tests are always welcome.

fson:
Could this be merged and a new release pushed to npm? It seems that most providers are broken at the moment.

PR #103, opened 2015-06-16
remove property username as it is not part of the google profile

qualquervalor:


geek:
Can you link to the google docs on this issue, I am having a difficult time tracking it down :/

qualquervalor:
This was motivated based on what actually get returned when making the call.  I'll see if I can find any docs to support the claim.

qualquervalor:
Try out their API explorer for this endpoint, there is an option to:
"Selector specifying which fields to include in a partial response. Use fields editor"
You can see that 'username' is not an available field.
https://developers.google.com/apis-explorer/#search/userinfo/m/oauth2/v2/oauth2.userinfo.get

qualquervalor:
Here is better documentation via their discovery documents:

The v2 endpoint
https://www.googleapis.com/discovery/v1/apis/oauth2/v2/rest

The original v1 endpoint that the code was using:
https://www.googleapis.com/discovery/v1/apis/oauth2/v1/rest

Here is the method call for userinfo:
"resources": {
  "userinfo": {
   "methods": {
    "get": {
     "id": "oauth2.userinfo.get",
     "path": "oauth2/v2/userinfo",
     "httpMethod": "GET",
     "response": {
      "$ref": "Userinfoplus"
     },

In both cases you see the response object is "Userinfoplus", does not include 'username'
"Userinfoplus": {
   "id": "Userinfoplus",
   "type": "object",
   "properties": {
    "email": {
     "type": "string",
     "description": "The user's email address."
    },
    "family_name": {
     "type": "string",
     "description": "The user's last name."
    },
    "gender": {
     "type": "string",
     "description": "The user's gender."
    },
    "given_name": {
     "type": "string",
     "description": "The user's first name."
    },
    "hd": {
     "type": "string",
     "description": "The hosted domain e.g. example.com if the user is Google apps user."
    },
    "id": {
     "type": "string",
     "description": "The obfuscated ID of the user."
    },
    "link": {
     "type": "string",
     "description": "URL of the profile page."
    },
    "locale": {
     "type": "string",
     "description": "The user's preferred locale."
    },
    "name": {
     "type": "string",
     "description": "The user's full name."
    },
    "picture": {
     "type": "string",
     "description": "URL of the user's picture image."
    },
    "verified_email": {
     "type": "boolean",
     "description": "Boolean flag which is true if the email address is verified. Always verified because we only return the user's primary email address.",
     "default": "true"
    }
   }
  }

PR #104, opened 2015-06-16
update the google oauth endpoint to use v2

qualquervalor:
I believe there is a v3 version but it is not yet documented, so the v2 version is what is current.


geek:
Any thoughts on using https://developers.google.com/+/web/api/rest/auth-migration#oauth2login instead?

qualquervalor:
[#108](https://github.com/hapijs/bell/pull/108) will supersede this and make this changes irrelevant

PR #106, opened 2015-06-17
authentication was breaking because of snake case 'settings.client_secret'  rather than 'settings.clientSecret'

rjmk:


gergoerdosi:
Duplicate of #101.

geek:
@rjmk thanks... it looks like this is fixed now.

PR #110, opened 2015-06-29
Fix google example

alanshaw:
This fixes the google example to use the `location` property not `redirect_uri` as this parameter is inferred from the current URL or the `location` setting if set. See https://github.com/hapijs/bell/blob/master/lib/oauth.js#L156

### DIFF
--- a/examples/google.js
+++ b/examples/google.js
@@ -18,12 +18,10 @@ server.register(Bell, function (err) {
         // Next, go to "APIs & auth >> Credentials and Create new Client ID
         // Select "web application" and set "AUTHORIZED JAVASCRIPT ORIGINS" and "AUTHORIZED REDIRECT URIS"
         // This will net you the clientId and the clientSecret needed.
-        // Also be sure to pass the redirect_uri as well. It must be in the list of "AUTHORIZED REDIRECT URIS"
+        // Also be sure to pass the location as well. It must be in the list of "AUTHORIZED REDIRECT URIS"
         clientId: '',
         clientSecret: '',
-        providerParams: {
-            redirect_uri: server.info.uri + '/bell/door'
-        }
+        location: server.info.uri
     });
 
     server.route({


qualquervalor:
I think it is more instructive that it continues to use the redirect_uri.  When you are using the Google Developer Console to setup credentials for your web app, you establish a client id, client secret and a redirect URI.  So, continuing to use redirect_uri at the same time as the clientId and clientSecret in the example feels intuitive.

alanshaw:
...but if you look at the code you'll see that whatever value you set for `redirect_uri` is completely discarded. You cannot influence the `redirect_uri` unless you use the `location` setting.

I think it's misleading to tell people to set a config parameter that isn't used.

In my use case I'm proxying from port 80 to 3000 using nginx - I was using the `redirect_uri` config option to set it to https://mydomain.com, but google was still trying to redirect to localhost:3000.

nelsonic:
:+1:

qualquervalor:
So, maybe the issue is that the redirect_uri is being overwritten.  The providerParams are read in https://github.com/hapijs/bell/blob/master/lib/oauth.js#L149 and then the query.redirect_uri is being overwritten on https://github.com/hapijs/bell/blob/master/lib/oauth.js#L156.  What if the line was change to check for the redirect_uri before overwriting it? 
```
query.redirect_uri =  query.redirect_uri  || internals.location(request, protocol, settings.location);
```

Also, If one example was going to be updated for this change, some of the others also use redirect_uri the same way and would need the same change.

alanshaw:
Whatever works best for you. I did consider doing that, but didn't know if it was by design that the `redirect_uri` was overwritten and normally prefer not to make a breaking change to the code if possible.

Updating the example seemed like a much less disruptive way of stopping others coming across the same problem. Happy to update the other examples if you like...also happy with the XOR solution.

PR #115, opened 2015-07-18
Add fields to the query

edimoldovan:
Added the fields which are then by default in the profile normalisation.

### DIFF
--- a/lib/providers/facebook.js
+++ b/lib/providers/facebook.js
@@ -20,7 +20,8 @@ exports = module.exports = function (options) {
         profile: function (credentials, params, get, callback) {
 
             var query = {
-                appsecret_proof: Crypto.createHmac('sha256', this.clientSecret).update(credentials.token).digest('hex')
+                appsecret_proof: Crypto.createHmac('sha256', this.clientSecret).update(credentials.token).digest('hex'),
+                fields: 'id,name,email,first_name,last_name,middle_name'
             };
 
             get('https://graph.facebook.com/v2.3/me', query, function (profile) {
--- a/test/oauth.js
+++ b/test/oauth.js
@@ -1382,7 +1382,7 @@ describe('Bell', function () {
 
                     Mock.override('https://graph.facebook.com/v2.3/me', function (uri) {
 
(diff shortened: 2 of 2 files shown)


edimoldovan:
Hi,

Can I get a little help with how to update the tests for this PR to pass the Travis build?

Thanks,
Eduárd

lib/providers/facebook.js
```diff
@@ -20,7 +20,8 @@ exports = module.exports = function (options) {
         profile: function (credentials, params, get, callback) {
 
             var query = {
-                appsecret_proof: Crypto.createHmac('sha256', this.clientSecret).update(credentials.token).digest('hex')
+                appsecret_proof: Crypto.createHmac('sha256', this.clientSecret).update(credentials.token).digest('hex'),
+                fields: 'id,name,email,first_name,last_name,middle_name'
```
edimoldovan: The graph query seams to return the fields only if they are explicitly provided.

ldesplat:
I am only getting the behavior you describe when using the v2.4 API but we're using v2.3 right now.

Either way, your PR looks good and works with 2.3.

For the tests, you need to change two lines in test/oauth.js:
```
expect(uri).to.equal('https://graph.facebook.com/v2.3/me?appsecret_proof=d32b1d35fd115c4a496e06fd8df67eed8057688b17140a2cef365cb235817102');
```
Change it to include the fields parameter
`&fields=id,name,email,...` and it should work.

I have to say that Lab/Expect are doing a really bad job at telling us what's wrong.

ldesplat:
@edimoldovan Is there anything more you need?

ldesplat:
> profileParams - an object of key-value pairs that specify additional URL query parameters to send with the profile request to the provider. The built-in facebook provider, for example, could have fields specified to determine the fields returned from the user's graph, which would then be available to you in the auth.credentials.profile.raw object.

If you still have problems before this is fixed, you can use profileParams as documented to get around it. Though I think those fields should definitely be included, especially as we upgrade to v2.4 API.

edimoldovan:
Hi @ldesplat 
Sorry, I missed your original reply, just noticed it now.

I tried the quick fix, doesn’t seam to work. 

How do I run tests locally?

Thanks,
Eduárd

ldesplat:
run `npm install` and then `npm test`

If you have issues and you don't know where it comes from you can also do `npm run test-cov-html`

edimoldovan:
@ldesplat If code/tests are fine with you, they’re ready to go from my side.

ldesplat:
@edimoldovan Thank you! :+1:

PR #126, opened 2015-08-17
Add provider credentialsHook function

rainabba:
...to provider settings to allow for manipulation of credentials before returned from Bell.

I've built a hybrid authentication system for Hapi 9 that provides 'basic', 'token' (using JWT) and 'social' auth. If the user is already authenticated with basic or token and they hit a route handled by Bell, I check to see if that social account is associated with the authenticated user and if not, I make the association. If they're not authenticated and they hit that route and the account is associated, I have a flow that will then use those credentials to associate them. To make this possible, I had to slightly modify behavior of bell and I distilled that down to a simple hook that can be passed with options which provides for a way to analyze and manipulate the credentials object with the request and provider name available. This change should have zero impact on existing users and only affect users who implement the hook via options and then it's very powerful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,10 @@ The `server.auth.strategy()` method requires the following strategy options:
 - `clientSecret` - the OAuth client secret (consumer secret).
 - `forceHttps` - A boolean indicating whether or not you want the redirect_uri to be forced to https. Useful if your hapi application runs as http, but is accessed through https.
 - `location` - Set the base redirect_uri manually if it cannot be inferred properly from server settings. Useful to override port, protocol, and host if proxied or forwarded.
+- `credentialsHook` - a function used to manipulate credentials just before they're returned for authentication. The function signature is `function( request, name, credentials )` where:
+    - `request` - the current request object being processed
+    - `name` - the name of the current bell provider being used (twitter, facebook, etc..)
+    - `credentials` - the present credentials object that you can modify before it's returned
 
 Each strategy accepts the following optional settings:
 - `cookie` - the name of the cookie used to manage the temporary state. Defaults to `'bell-provider'` where 'provider' is the provider name
--- a/lib/index.js
+++ b/lib/index.js
@@ -60,7 +60,8 @@ internals.schema = Joi.object({
     config: Joi.object(),
     profileParams: Joi.object(),
     forceHttps: Joi.boolean().optional().default(false),
-    location: Joi.string().optional().default(false)
+    location: Joi.string().optional().default(false),
+    credentialsHook: Joi.func().optional()
 });
 
 
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -120,6 +120,7 @@ exports.v1 = function (settings) {
 
             settings.provider.profile.call(settings, credentials, payload, get, function () {
 
+                if ( settings.credentialsHook ) { credentials = settings.credentialsHook( request, name, credentials ); }
                 return reply.continue({ credentials: credentials });
             });
         });
@@ -279,6 +280,7 @@ exports.v2 = function (settings) {
 
             settings.provider.profile.call(settings, credentials, payload, get, function () {
 
+                if ( settings.credentialsHook ) { credentials = settings.credentialsHook( request, name, credentials ); }
                 return reply.continue({ credentials: credentials });
             });
         });


ldesplat:
Thank you very much for submitting this PR. It is very appreciated.

Could you not handle this logic in the route or pre handler for the bell endpoint? Or even do it with the `onPostAuth` hook of hapi?

Your PR also lacks tests and does not conform to the code guidelines but let's make sure we agree on the concept of it before you fix that.

rainabba:
Sorry for the lack of convention. I'm still pretty new to Github and testing to be honest; decades in a .net shop and I'm trying to learn good, new habits :)

Using onPostAuth won't work because the significance of this change is during auth, between strategies. I'm implementing multiple strategies on single routes and this works because of the order those strategies are applied. Put another way, this is needed in order for me to make the strategies work together as I require. To help clarify and add context, I'm including my actual hook function below.

```
bellCredentialsHook: function( request, name, credentials ) {
    var newCredentials = {};

    if ( request.state.tfTqE27ZyBgG ) {
        newCredentials = request.state.myUniqueCookieName;
        newCredentials.socialProvider = { name: name, credentials: credentials };    
    } else {
        newCredentials = credentials;
        newCredentials.socialProvider = { name: name, credentials: credentials };
    }

    return  newCredentials;
}
```

Without this, I loose the auth context created by my JWT and hapi-auth-cookie strategies that existed prior to bell processing.

As for the testing, I did include the JOI validation of the function. I'm not sure how to test further since the actual function doesn't exist as part of Bell, but then I'm still fuzzy on automated testing in general.

Could you direct me to the code-guidelines I missed? I know it's not your responsibility, but I would like to be able to contribute by the rules and this process isn't clear nor have I found a clear, concise guide that doesn't assume other knowledge outside of writing code and using git.

ldesplat:
The style guide: http://hapijs.com/styleguide

You can run the tests included with bell by doing `npm test` (after `npm install`). You can also do `npm run test-cov-html` and it will create a html file that shows graphically where there is coverage missing. Bell currently has 100% coverage so any change in coverage comes from any changes you make. In `/test/oauth.js` you will need to add tests that cover your use case. We use [hapijs/lab](https://github.com/hapijs/lab) and [hapijs/code](https://github.com/hapijs/code) to run tests and make assertions.

Also, running npm test will check for a subset of style issues.

All this is true for the [hapi ecosystem](http://hapijs.com/contribute).

---

I need some time to understand the full problem still. So, you create one auth strategy that combines all of them together? Bell is usually involved at only one endpoint per provider and implementations tend to redirect away from it as soon as it did its job. I still do not see why it cannot be handled further down the chain. I need to research this a bit. @geek or @hueniverse Do you have any input you could give?

rainabba:
@ldesplat I will review the tests and hopefully from the examples, figure out how to get my changes covered :)

I wouldn't say that I've created a new strategy that combines the others, just implemented all 3 in a particular way. The routes that bell catches are setup so that they do as normal and the handler for those routes users the provided credentials to either associate credentials with the presently authenticated user (this is where it gets tricky) or uses the credentials to validate the user as registered and then sets up a session if they are and one doesn't already exist.

In the case where the user is already authenticated (either using my basic method or JWT), then Bell will setup request.auth in a way that I have no way of knowing the user was already authenticated and what their previous credentials were. All my hook does ultimately is ensure that I have the information needed in the new credentials.

I'm curious why the specifics of my implementation are much concern. The approach I took (the PR submitted) was thought out heavily to ensure that it would have zero impact on anyone that doesn't explicitly use it, but allows for lots of control when needed, or is this just a matter of curiosity?

I think I mentioned elsewhere the other day, but this really comes down to the fact that I don't need Bell as an auth strategy at all, but really I just need the incredibly well-done OATH providers. If there was a project that let me put this oauth flow on a route then let me handle the result (without tampering with the auth object at all), then I wouldn't even need this, but this hook effectively makes that behavior possible because I could overwrite the credentials object with request.auth in the hooked function.

ldesplat:
@rainabba I appreciate the zero impact. Believe me, I do. I am not trying to give you a hard time either :) I just want to make sure the added complexity (more options) which we have to maintain is worth it. It may or may not be, I am just trying to establish that fact!

You are actually in luck. We do expose it just the way you like :)

You see `server.expose('oauth', OAuth);` at https://github.com/hapijs/bell/blob/master/lib/index.js#L26

Therefore, you can now do `server.plugins.bell.oauth` and call the `v2` method with your settings. To pass your settings, you'll have to require your provider (custom or from `bell/lib/providers`) and merge it. The schema is https://github.com/hapijs/bell/blob/master/lib/index.js#L35-L63.

Cheers,
Lois.

ldesplat:
Closing this. If you still need this, reopen a PR with tests but hopefully I have given you a working alternative.

PR #129, opened 2015-08-20
added allowRuntimeProviderParams to allow runtime query params 

### LINKED ISSUE #125: Provide variable providerParams
My provider accepts extra query params on the auth route, which you can set with `providerParams`. But afaik I don't see an option to change these values on every request. For example I can provide a `login_hint` query param but this hint is different on every request

AdriVanHoudt:
to be send to auth url // allows runtime changing query params // fixes #125

we can discuss naming ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ Each strategy accepts the following optional settings:
     - Facebook supports `display` ('page', 'popup', or 'touch'), `auth_type`, `auth_nonce`.
     - Google supports `access_type`, `approval_prompt`, `prompt`, `login_hint`, `user_id`, `hd`.
     - Twitter supports `force_login`, `screen_name`.
+- `allowRuntimeProviderParams` - allows passing query parameters from a **bell** protected endpoint to the auth request. It will merge the query params you pass along with the providerParams and any other predefined ones. Be aware that this will override predefined query parameters! Default to `false`.
 - `scope` - Each built-in vendor comes with the required scope for basic profile information. Use `scope` to specify a different scope
   as required by your application. Consult the provider for their specific supported scopes.
 - `config` - a configuration object used to customize the provider settings. The built-in `'twitter'` provider accepts the `extendedProfile`
--- a/lib/index.js
+++ b/lib/index.js
@@ -55,6 +55,7 @@ internals.schema = Joi.object({
     ttl: Joi.number(),
     domain: Joi.string().allow(null),
     providerParams: Joi.object(),
+    allowRuntimeProviderParams: Joi.boolean().default(false),
     scope: Joi.array().items(Joi.string()).when('provider.protocol', { is: 'oauth2', otherwise: Joi.forbidden() }),
     name: Joi.string().required(),
     config: Joi.object(),
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -60,6 +60,10 @@ exports.v1 = function (settings) {
                 var authQuery = settings.providerParams ? Hoek.clone(settings.providerParams) : {};
                 authQuery.oauth_token = payload.oauth_token;
 
+                if (settings.allowRuntimeProviderParams ) {
+                    Hoek.merge(authQuery, request.query);
+                }
+
                 return reply.redirect(settings.provider.auth + '?' + internals.queryString(authQuery));
             });
         }
@@ -147,10 +151,12 @@ exports.v2 = function (settings) {
         if (!request.query.code) {
             var nonce = Cryptiles.randomString(22);
             query = Hoek.clone(settings.providerParams) || {};
-            protocol = request.connection.info.protocol;
-            if (settings.forceHttps) {
-                protocol = 'https';
+
+            if (settings.allowRuntimeProviderParams ) {
+                Hoek.merge(query, request.query);
             }
+
+            protocol = settings.forceHttps ? 'https' : request.connection.info.protocol;
             query.client_id = settings.clientId;
             query.response_type = 'code';
             query.redirect_uri = internals.location(request, protocol, settings.location);
--- a/test/oauth.js
+++ b/test/oauth.js
@@ -7,6 +7,7 @@ var Code = require('code');
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Lab = require('lab');
+var Joi = require('joi');
 var Mock = require('./mock');
 var OAuth = require('../lib/oauth');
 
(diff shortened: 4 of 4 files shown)


AdriVanHoudt:
2 things:
* tests already failed with a clean clone (not sure why it is failing)
* why not just use the built in lint config of Lab?

AdriVanHoudt:
so any thoughts on this?

lib/oauth.js
```diff
@@ -147,27 +148,34 @@ exports.v2 = function (settings) {
...
-            query.response_type = 'code';
-            query.redirect_uri = internals.location(request, protocol, settings.location);
-            query.state = nonce;
 
-            var scope = settings.scope || settings.provider.scope;
-            if (scope) {
-                query.scope = scope.join(settings.provider.scopeSeparator || ' ');
-            }
+            // Validate runtime query params
+            return Joi.validate(request.query, settings.runtimeQuerySchema, function (err, value) {
```
ldesplat: I know it's the default but can you explicitly state { allowUnknown: false } in the Joi.validate options.

ldesplat:
@AdriVanHoudt It looks good to me. I'll try tomorrow with the google provider. Please provide some documentation :)

@hueniverse Can you please review this? I really want to make sure we're not introducing a security issue here. My main worry is that we're not verifying the values sent are what the application wanted to send. I don't think it's an issue here but still would like an expert to review this!

lib/oauth.js
AdriVanHoudt: will do

AdriVanHoudt:
@ldesplat ok, was holding off until i was sure it was good

AdriVanHoudt:
adding allowUnknown seems to throw for some reason @Marsup ?

ldesplat:
Well I am dumb sometimes. Even if we verify the content it does not matter because the client is redirected so an attacker can do whatever they want with it.

I wonder if your original idea of just hoek.merge all the world is the right way to do it. The checks I've made you do are useless...

ldesplat:
@AdriVanHoudt You need to put the options after the schema.

AdriVanHoudt:
derp :P updated now

ldesplat:
I'll merge this (with your documentation) after the weekend if nothing bad comes up (in comments here or my testing).

lib/oauth.js
```diff
@@ -459,9 +467,9 @@ internals.Client.prototype.signature = function (method, baseUri, params, oauth,
     normalized.sort(function (a, b) {
 
         return (a[0] < b[0] ? -1
-                            : (a[0] > b[0] ? 1
-                                           : (a[1] < b[1] ? -1
-                                                          : (a[1] > b[1] ? 1 : 0))));
+                : (a[0] > b[0] ? 1
+                   : (a[1] < b[1] ? -1
+                      : (a[1] > b[1] ? 1 : 0))));
     });
```
hueniverse: More readable the way it was.

lib/oauth.js
```diff
@@ -472,8 +480,8 @@ internals.Client.prototype.signature = function (method, baseUri, params, oauth,
     // String Construction (3.4.1.1)
 
     var baseString = internals.encode(method.toUpperCase()) + '&' +
-                     internals.encode(baseUri) + '&' +
-                     internals.encode(normalizedParam);
+        internals.encode(baseUri) + '&' +
+        internals.encode(normalizedParam);
```
hueniverse: More readable the way it was.

lib/oauth.js
```diff
@@ -147,27 +148,34 @@ exports.v2 = function (settings) {
...
-            query.response_type = 'code';
-            query.redirect_uri = internals.location(request, protocol, settings.location);
-            query.state = nonce;
 
-            var scope = settings.scope || settings.provider.scope;
-            if (scope) {
-                query.scope = scope.join(settings.provider.scopeSeparator || ' ');
-            }
+            // Validate runtime query params
+            return Joi.validate(request.query, settings.runtimeQuerySchema, { allowUnknown: false }, function (err, value) {
```
hueniverse: Rework this without using the validate callback (use the returned value format instead). It's easier to read.

hueniverse:
I am not following the requirements here. Why set up a joi schema? I understand supporting per-request query additions but this feels complex for no reason. Are you then going to document every provider's additional query params?!

ldesplat:
@hueniverse Thanks for reviewing it.

I did not want a user attempt to use the login endpoint with extra parameters that we may not want to pass along. But we can't protect against that obviously since the whole oauth flow is not designed so.

ie. I am not used to taking query params and passing them along with no verification. But in this case it makes sense.

@AdriVanHoudt I am sorry, you can remove the verification of parameters.

AdriVanHoudt:
no problem, will make the changes according to the comments
Also the indent thing is just me indenting the whole file the same (habit, will change it back)

AdriVanHoudt:
should be done now, let me know your feedback and if it is ok I will squash for you ;)

README.md
```diff
@@ -113,6 +113,7 @@ Each strategy accepts the following optional settings:
     - Facebook supports `display` ('page', 'popup', or 'touch'), `auth_type`, `auth_nonce`.
     - Google supports `access_type`, `approval_prompt`, `prompt`, `login_hint`, `user_id`, `hd`.
     - Twitter supports `force_login`, `screen_name`.
+- `allowRuntimeQueryParams` - will allow you to pass along query params to the auth request that change on every request. Bassicly merging the query params you pass along with the predefined ones. Be aware that this will override predefined query params! Defaults to `false`.
```
hueniverse: Should be edited to be more consistent with the other text. E.g. `'allows passing query parameters...'`

lib/index.js
```diff
@@ -55,6 +55,7 @@ internals.schema = Joi.object({
     ttl: Joi.number(),
     domain: Joi.string().allow(null),
     providerParams: Joi.object(),
+    allowRuntimeQueryParams: Joi.boolean().default(false),
```
ldesplat: We've only implemented it for oauth2 so adding `.when('provider.protocol', { is: 'oauth2', otherwise: Joi.forbidden() })` might be useful so people don't wonder why it is not working.

ldesplat: Renaming it to `allowRuntimeProviderParams` might make the documentation easier since they are related.

README.md
```diff
@@ -113,6 +113,7 @@ Each strategy accepts the following optional settings:
     - Facebook supports `display` ('page', 'popup', or 'touch'), `auth_type`, `auth_nonce`.
     - Google supports `access_type`, `approval_prompt`, `prompt`, `login_hint`, `user_id`, `hd`.
     - Twitter supports `force_login`, `screen_name`.
+- `allowRuntimeQueryParams` - allows passing query parameters to the auth request by adding them to the first call to the endpoint. It will merge the query params you pass along with the predefined ones. Be aware that this will override predefined query params! Defaults to `false`.
```
ldesplat: Maybe?
A boolean that allows passing query parameters from a bell protected endpoint to the auth request. It will merge the query params you pass along with the providerParams and any other predefined ones. Be aware that this will override all other query parameters! Default to \`false\`.

ldesplat:
Thanks! This looks great. Tried it out with the google provider and works great. I am ready to merge when you squash it :)

lib/index.js
AdriVanHoudt: do you also want it for oath1? (not used it so not sure if it applies)

AdriVanHoudt:
Ok squashed and allowed on oauth1

ldesplat:
You forced me to read the oauth1 spec finally!

Looks like it's allowed though I have not found any providers allowing extra parameters just yet. It looks good so I am merging it!

Thank you so very much for this PR and bearing with me!

ldesplat:
It's in the 5.1.0 release.

AdriVanHoudt:
haha sorry :stuck_out_tongue_winking_eye: @hueniverse will be happy (as far as I recall he helped write it)

And no problem, it was my pleasure

PR #130, opened 2015-08-20
fixed linting issues // now extends lab linting

AdriVanHoudt:
I added special rule def in test files because it was used so much that fixing it would take a lot of time and the rule is more a precaution anyway, so be carefull with you res's in your tests kids :exclamation:


ldesplat:
Thanks, I'll merge this once I figure out what is wrong with our tests. The last merged PR passed no problem :(

AdriVanHoudt:
maybe a dep patch update or something?

AdriVanHoudt:
Will rebase soon

ldesplat:
No need it does it automatically when it can.

AdriVanHoudt:
:+1:

PR #131, opened 2015-08-20
Fix wrong tests. OAuth1 does not include expiresIn for yahoo/bitbucke…

ldesplat:
…t/twitter. Remove circular reference.

The tests were broken for a long time but upgraded dependencies and hoek > 2.11 uncovered those issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "boom": "2.x.x",
     "cryptiles": "2.x.x",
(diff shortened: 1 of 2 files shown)


ldesplat:
Travis fails because of linting issues which are partially fixed in #130

AdriVanHoudt:
Nice was thinking about doing a dep update but I was waiting for the tests to be fixed :D

PR #137, opened 2015-09-09
Node v4

ldesplat:


Marsup:
Don't forget the sudo false to take advantage of new travis infrastructure.

ldesplat:
Thank you, did not know :)

.travis.yml
```diff
@@ -2,5 +2,4 @@ language: node_js
 
 node_js:
   - 0.10
-  - 0.12
```
jalcine: Why not still support `0.12`?

ldesplat: Hapi never really supported `0.12`. 
See https://github.com/hapijs/discuss/issues/85
> We will not support 0.12 as there is no point. Moving to 0.12 is not worth the effort from 0.10 which is the most stable, battle tested version. As for io.js, once v4.0 is out and is the new official version of node.js, we will start testing against it. Beyond this, it is hard to say.

jalcine: [**@ldesplat**](https://github.com/ldesplat) commented on [Sep 9, 2015, 5:27 PM EDT](https://github.com/hapijs/bell/pull/137#discussion_r39098740 "2015-09-09T21:27:01Z - Replied by Github Comment Enhancer"):
> Hapi never really supported `0.12`.   
> 
> See [hapijs/discuss#85](https://github.com/hapijs/discuss/issues/85 "io.js compatibility and support")
> 
> > We will not support 0.12 as there is no point. Moving to 0.12 is not worth the effort from 0.10 which is the most stable, battle tested version. As for io.js, once v4.0 is out and is the new official version of node.js, we will start testing against it. Beyond this, it is hard to say.

Good to know!

PR #138, opened 2015-09-10
Fix protocol detection - Resolves #136

### LINKED ISSUE #136: OAuth provider is detecting 'socket' instead of 'http' as protocol
Consider the following OAuth URL generated by Bell: 

https://www.facebook.com/v2.3/dialog/oauth?client_id=859994797418065&response_type=code&redirect_uri=socket%3A%2F%2Fwebtier.undergroundcellar.com%2Fsignin%2Ffacebook&state=y2uYfM8i97sgrzcAQnn7w6&scope=email

Looking closer, at `redirect_uri` query param, we see the protocol is "socket"

This is coming from here: https://www.facebook.com/v2.3/dialog/oauth?client_id=859994797418065&response_type=code&redirect_uri=socket%3A%2F%2Fwebtier.undergroundcellar.com%2Fsignin%2Ffacebook&state=y2uYfM8i97sgrzcAQnn7w6&scope=email, namely `protocol = request.connection.info.protocol;` 

Does anyone know why this is returning "socket" instead of "http" (here https is not being used). I was thinking of adding this after L153 but it feels like the wrong approach: 

```js
            if (protocol === 'socket') {
                protocol = 'http';
            }
```

Any advice?

Thanks,
Ben
fyi @d-alejo

bherila:
I have this working on http://webtier.undergroundcellar.com to test... but I have not yet figured out how to get the test working as 'protocol' does not seem allowed to be set.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -24,6 +24,9 @@ exports.v1 = function (settings) {
         var name = settings.name;
         var query = request.query;
         var protocol = request.connection.info.protocol;
+        if (protocol === 'socket') {
+            protocol = 'http';
+        }
 
         // Bail if the upstream service returns an error
 
@@ -123,6 +126,7 @@ exports.v1 = function (settings) {
             };
 
             settings.provider.profile.call(settings, credentials, payload, get, function () {
+
                 return reply.continue({ credentials: credentials });
             });
         });
@@ -155,7 +159,14 @@ exports.v2 = function (settings) {
                 Hoek.merge(query, request.query);
             }
 
-            protocol = settings.forceHttps ? 'https' : request.connection.info.protocol;
+            protocol = request.connection.info.protocol;
+            if (settings.forceHttps) {
+                protocol = 'https';
+            }
+            else if (protocol === 'socket') {
+                protocol = 'http';
+            }
+
             query.client_id = settings.clientId;
             query.response_type = 'code';
             query.redirect_uri = internals.location(request, protocol, settings.location);
@@ -191,6 +202,9 @@ exports.v2 = function (settings) {
         protocol = request.connection.info.protocol;
         if (settings.forceHttps) {
             protocol = 'https';
+        }
+        else if (protocol === 'socket') {
+            protocol = 'http';
         }
 
         query = {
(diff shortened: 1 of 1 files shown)


ldesplat:
The question I have with this is, how do we know to use http over https. Obviously for your use case, that's what you want but I would hate to make that the default.

And now, we get into the problems of knowing which protocol to redirect to. It is a semi-standard to use a header like `x-forwarded-proto` for a reverse proxy to tell the application which protocol was used to access the resource.

I don't think it is something that bell can just solve by checking those headers since everybody has slightly different setups. So it's sounding like we need some sort of configuration where people can set this up. 
Could solve it like so:
1. `redirectProtocol` forces the protocol to be this string
2. `findProtocol` custom function that looks at the request and returns the protocol
3. `protocolHeader` just a header we look at to set the protocol
4. `forceHttp` force http protocol

I am leaning towards `3` or `4` but I have to think about it some more.

Or you know, you could just use https and set `forceHttps` right now :)

For the test cases:
You'll have to set up a scenario where you create a Hapi server using a `socket` protocol and verify that it redirects to the protocol you expect. Can just copy a successful testcase and actually start a server.

bherila:
I agree we can't rely on the headers necessarily. This particular test application is a Node site running on Windows Azure. Instead of a setting `forceHttp`, maybe it would be best to have a more generic setting `forceProtocol` which could be http, https, etc. ? Although admittedly, it would be redundant against `forceHttps`. 

BTW, I would really love to use HTTPS however this is a development environment and our domain uses an EV certificate which does not have the development subdomain listed. So I'd have to get the EV certificate redone which is a big project. In production, it will be using HTTPS so this problem won't exist. 

I'm really a big noob when it comes to Hapi, and running up against a deadline so I am probably going to have to use my untested but working fork for the time being. But I would like to get this done the right way and be helpful to the community (surely someone else will run into this issue eventually) so hopefully I'll have some time this weekend to play around with the test cases.

ldesplat:
Well looks like there is `location` already.

```location - Set the base redirect_uri manually if it cannot be inferred properly from server settings. Useful to override port, protocol, and host if proxied or forwarded.```

bherila:
Oh great, I didn't notice that before! Thanks :+1:

Issue #127, opened 2015-08-17
Document Bell Core API and Providers API


ldesplat:
- Don't forget we expose the OAuth implementation so it can be used without being part of the default bell scheme.

ldesplat:
Fixed by #139

Issue #109, opened 2015-06-22
Provider wiki


qualquervalor:
It would be useful if there were a wiki that provided reference material associated with the various providers.  It can be somewhat difficult to track down the documentation about endpoints, scopes, response bodies, etc..  This information has to be researched when creating a provider for the first time, so it couldn't hurt if it was recorded at that time.  This should help in the long run in maintaining the code when the provider updates there service.  Here is a sample associated with the Google provider:

Endpoints:
Goggle discovery doc for the plus service:
https://www.googleapis.com/discovery/v1/apis/plus/v1/rest

Auth and Token Details:
Auth and Token based on values found in the OpenIDConnect discovery document:
https://accounts.google.com/.well-known/openid-configuration

Scopes:
Scope discussion can be found here:
https://developers.google.com/+/web/api/rest/oauth#plus.login

Response body:
Documentation of response body:
https://developers.google.com/+/web/api/rest/latest/people#resource

ldesplat:
What do you think about documenting the provider details directly in the source code of the providers?

It would have more of a chance at staying up to date...

qualquervalor:
I think it would still merit having a significant portion if it outside the code base.  It is easier to diff code changes version to version if there is not a large portion of changing comments sprinkled throughout a file.  Plus some of this type of information could be hard to find if it was dispersed throughout a code base.  Regardless of whether it is put in the code or in a wiki, it will eventually get dated.  If it was in the code would you then not want to accept a pull request if a volunteer submitted code fixes but didn't update the documentation.

ldesplat:
To directly answer your question, I would not accept the PR as is. That's what is done on other Hapi repos.

Now, I would love to better document each provider from a user's perspective. I think that this is not documented currently (spread around the README.md and I am not sure if it is exhaustive). So document, exactly what type of options you can pass to it in a specific section of the README or its own markdown file.

Having a link to the specifics of each implementations, facebook docs, google docs, phabricator docs, ... I could see that being put with that documentation. But I would not summarize that information in here.

Would you be alright with such a solution or is there something I am missing?

qualquervalor:
More documentation in any format would be good, so yes.

ldesplat:
Fixed by #139

PR #139, opened 2015-09-11
Split up the documentation and add provider specific information plus…

ldesplat:
… some more details

- README.md much like Hapi's
- General API documentation
  - Added a section on how to use without a strategy
- Providers documentation
  - What does each provider take as config
  - What does each provider return in their profile information
  - What is the `scope` that each provider has on by default


ldesplat:
@qualquervalor Does this solve #109 ?

Also satisfies #127

qualquervalor:
It looks good, having links to the providers documentation is nice.  I would suggest also calling out the endpoints, auth and token details for each provider, in Provider.md

PR #158, opened 2015-10-30
[github] if github doesn't like auth parameters

mrose17:
it does set `query.error` (but not to `"access_denied"`), it also sets `query.error_description` ... which is what this PR looks for. i think a more general solution is to have an optional function that looks at `request.query` and returns either `null` or `boom.whatever()` ... but, that's outside of my "pay grade"

thanks for the wonderful package!

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -141,7 +141,7 @@ exports.v2 = function (settings) {
         var state;
 
         // Bail if the upstream service returns an error
-        if (request.query.error === 'access_denied' || request.query.denied || query.error_description) {
+        if (request.query.error === 'access_denied' || request.query.denied || request.query.error_description) {
             return reply(Boom.internal('App was rejected'));
         }
 
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -25,7 +25,6 @@ exports.v1 = function (settings) {
         var query = request.query;
         var protocol = request.connection.info.protocol;
 
-console.log('0: ' + require('util').inspect(query, { depth: null }));
         // Bail if the upstream service returns an error
 
         if (query.error === 'access_denied' || query.error_description || query.denied) {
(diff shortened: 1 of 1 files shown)


lib/oauth.js
```diff
@@ -142,7 +141,7 @@ exports.v2 = function (settings) {
         var state;
 
         // Bail if the upstream service returns an error
-        if (request.query.error === 'access_denied' || request.query.denied) {
+        if (request.query.error === 'access_denied' || request.query.denied || query.error_description) {
```
ldesplat: This should be `request.query.error_description`

ldesplat:
You can see in the spec http://tools.ietf.org/html/rfc6749#section-4.1.2.1 that request.query.error is mandatory.

This PR changes behavior since this piece of code is only checking for `access_denied` when there are a few other error conditions that can exist. With your code change, it would go in the if statement even when other errors are present. I don't know why it only checks for `access_denied`. It does not make any sense to me right now.

Do you have a special scenario? Should we just be checking for the existence of `request.query.error`?

mrose17:
i agree that request.query.error is mandatory! i think the issue is that the original code tested for a specific value (access_denied) but github is returning something else.  this is perhaps the core issue.

what github is complaining about is the `uri_redirect` parameter sent by the code not matching the registered value.  so, perhaps better code is something like:

        if (request.query.error === 'access_denied' || request.query.denied) {
            return reply(Boom.forbidden('App was rejected'));
        }
        if (request.query.error) {
            return reply(Boom.internal(request.query.error));
        }

your thoughts?

thanks!

ldesplat:
I'll go along. You might as well include `error_description` and `error_uri` also. You will need to add some tests to get to 100% code coverage. Also, I have to investigate a special case when the user has the state cookie set already, I don't want that to linger. We should do `.unstate` if it's present.

Can you do something like that?

You can run `npm run test-cov-html` to see a graphical representation of the code coverage.

Thank you!

mrose17:
hah! you are asking me to do *real work*... ok, it will take me a few days to free up the time, and then i'll submit another PR... thanks!

ldesplat:
Fixed in 705303f

PR #162, opened 2015-11-03
Add error if using non-https server when isSecure is set to true

tanepiper:
Today I've spent most of my morning digging around the Hapi internals to find why the state cookie was being destroyed, only to find that it was because I'm developing locally using http, but had not set `isSecure: false` in the options.

The issue is that the error about the cookie not being set at [line 77](https://github.com/hapijs/bell#/L77) isn't too helpful.  Took me into Hapi's own request internals because I was seeing on the outgoing request when I did console.log

```
{ 'bell-twitter':
   { name: 'bell-twitter',
     value:
      { token: '<token>',
        secret: '<secret>',
        query: {} } } }
Debug: internal, implementation, error
    TypeError: Cannot read property 'cookies' of undefined
    at [object Object].internals.Request.internals.Request._execute.internals.Request._lifecycle.internals.Request._invoke.internals.Request._reply.internals.Request._finalize (D:\Documents\GitHub\myapp\node_modules\hapi\lib\request.js:473:28)
```

Then on the request back from twitter:

```
{ undefined: '' }
{}
```

It was quite unhelpful.  Hopefully making this message explicit will make it easier for other people to avoid the same mistake

### DIFF
--- a/API.md
+++ b/API.md
@@ -29,7 +29,7 @@ server.register(require('bell'), function (err) {
         password: 'cookie_encryption_password',
         clientId: 'my_twitter_client_id',
         clientSecret: 'my_twitter_client_secret',
-        isSecure: false     // Terrible idea but required if not using HTTPS
+        isSecure: false     // Terrible idea but required if not using HTTPS especially if developing locally
     });
 
     // Use the 'twitter' authentication strategy to protect the
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ server.register(require('bell'), function (err) {
         provider: 'twitter',
         password: 'cookie_encryption_password',
         clientId: 'my_twitter_client_id',
-        clientSecret: 'my_twitter_client_secret'
+        clientSecret: 'my_twitter_client_secret',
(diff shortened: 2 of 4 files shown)


tanepiper:
I was able to get the first test passing, but not the one for V2.  This will check for forceHttps being set.

ldesplat:
I like better error messages for the developer :+1: . You will also have to add `!settings.location` since you can redefine the protocol using that feature too. It's a bit of a mess of features.

Thank you for adding tests too. Do you mind if this eventually gets merged for node 4+, Hapi 11?

tanepiper:
@ldesplat OK - re-written to check for the correct protocol, and tests written to confirm it works in call cases (forceHttps, location and not set).  Happy for this to be merged in ASAP.

tanepiper:
@ldesplat OK I've rebased the code to be ES6 and up to date with the master branch here - but it doesn't seem to be happy with the tests on the CI, but they seem to all run locally fine

ldesplat:
Top notch contribution. Thank you so much.

ldesplat:
It is now included in the 6.0.0 release that's released on NPM.

PR #169, opened 2015-12-02
vk provider updated. Email field added to profile object.

antipin:
In case of VK provider we can get users email on https://oauth.vk.com/access_token request only. So we should pass it to profile object.


ldesplat:
Was this a mistake or do we need to update the vk provider for anything?

antipin:
I think we need this update, because email is important field i guess, but my changes leads to test failure, so I revoke it

PR #173, opened 2015-12-29
Added tutorial 'Social Login with Twitter using hapi.js' to README

p-meier:


gergoerdosi:
I think hapijs.com would be a better place to link to this tutorial:

https://github.com/hapijs/hapijs.com/blob/master/lib/resources.js#L86

ldesplat:
I think linking it in hapijs.com is very important too. It's nice for people to have it here as well since Bell and OAuth are not very friendly concepts to newcomers. It's up to the maintainer (currently me) to ensure that if it gets outdated to remove the reference here.

p-meier:
Thanks for merging. I will add it to hapijs.com too...

PR #174, opened 2015-12-30
Update bitbucket to use the oAuth2 endpoint

mattapperson:


ldesplat:
Very nice!

Two things:
1. The tests are failing so ensure you update them. In this case I think they just need to be told to expect oauth2 mocks.
bitbucket
  ✖63) authenticates with mock
  ✖64) authenticates with mock (last_name is empty)
2. Please also update the documentation in Providers.md
https://github.com/hapijs/bell/blob/master/Providers.md#bitbucket

Thanks!

ldesplat:
Thank you very much. You should have put a comment, I completely missed this!

PR #179, opened 2016-02-01
Added Twitch provider

Belar:
Provider for [Twitch](http://www.twitch.tv/) with test and example.


ldesplat:
Thank you so much!

Belar:
Thank you for the plugin, happy I could contribute.

PR #183, opened 2016-02-07
skipProfile option

bbatliner:
There are some cases when I want to use bell with OAuth scopes that don't include the profile. For example, I'm using the Google Calendar scope, but not the profile scope, so bell wouldn't be able to fetch the profile. This lets the user skip fetching the profile from their provider config, without having to write a new provider with their blank `profile` function.


ldesplat:
So, some providers give you the option of `extendedProfile: true/false`. Feels like the Google one (and really all of them) should definitely implement that. 

Now I think I also understand the use case for this but let me fully understand the impact. I don't like how there are no additional tests and somehow you are getting 100% test coverage.. something is not right.

ldesplat:
Also, need docs for this PR in API.md :)

bbatliner:
Hey @ldesplat, thanks for the feedback. I updated the API and added a couple tests for oauth v1 and v2 with skipProfile. Let me know if there's anything else I can do!

ldesplat:
Thanks! Will go through it today.

ldesplat:
It is now in 6.3.0. Thank you :)

PR #185, opened 2016-02-07
implementation of microsoftonline.com oauth2,

pavelmickevic:
Documentation follows: [OAuth 2.0 in Azure AD](https://msdn.microsoft.com/en-us/library/azure/dn645545.aspx)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -45,6 +45,10 @@ internals.schema = Joi.object({
         token: Joi.string().required(),
         headers: Joi.object(),
         profile: Joi.func(),
+        resource: Joi.string().when('name', { is: 'microsoftonline', then: Joi.required(), otherwise: Joi.forbidden() }),
+        response_type: Joi.string().optional().default('code'),
+        response_mode: Joi.string().optional().default('query'),
+        nonce: Joi.boolean().optional().default(false),
         scope: Joi.array().items(Joi.string()).when('protocol', { is: 'oauth2', otherwise: Joi.forbidden() }),
         scopeSeparator: Joi.string().when('protocol', { is: 'oauth2', otherwise: Joi.forbidden() }),
         version: Joi.string()
--- a/lib/index.js
+++ b/lib/index.js
@@ -45,7 +45,7 @@ internals.schema = Joi.object({
         token: Joi.string().required(),
         headers: Joi.object(),
         profile: Joi.func(),
-        resource: Joi.string().when('name', { is: 'microsoftonline', then: Joi.required(), otherwise: Joi.forbidden() }),
+        resource: Joi.string().when('name', { is: 'microsoftonline', then: Joi.required(), otherwise: Joi.optional() }),
         response_type: Joi.string().optional().default('code'),
         response_mode: Joi.string().optional().default('query'),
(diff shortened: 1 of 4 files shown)


lib/providers/microsoftonline.js
```diff
@@ -0,0 +1,52 @@
...
+        response_mode: 'query',
+        response_type : 'code id_token',
+        resource: 'https://graph.windows.net',
+        profile: function (credentials, params, get, callback) {
+
+            const query = {
+                'api-version': '1.6'
+            };
+
+            const me = ('https://graph.windows.net/{tenantId}/me').replace('{tenantId}', tenantId);
```
ldesplat: Just use ES6 template string https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings

ldesplat:
Before, I make any more comments.. is there any other documentation page for their implementation & usage of oauth2 & openid?

lib/oauth.js
```diff
@@ -167,9 +167,17 @@ exports.v2 = function (settings) {
             }
 
             query.client_id = settings.clientId;
-            query.response_type = 'code';
+            query.response_type = settings.provider.response_type;
             query.redirect_uri = internals.location(request, protocol, settings.location);
             query.state = nonce;
+            if (settings.provider.nonce) {
+                query.nonce = nonce; // required by microsoftonline
```
ldesplat: In this case we need to generate a new nonce I am pretty sure.

lib/index.js
```diff
@@ -45,6 +45,10 @@ internals.schema = Joi.object({
         token: Joi.string().required(),
         headers: Joi.object(),
         profile: Joi.func(),
+        resource: Joi.string().when('name', { is: 'microsoftonline', then: Joi.required(), otherwise: Joi.forbidden() }),
```
ldesplat: `otherwise: Joi.optional()` . Someone may create their own custom provider that needs this but not necessarily contribute back to this project.

lib/oauth.js
pavelmickevic: For me it looks like a rudimental property left by MS. It does not matter, what nonce would be sent, they return `state` property anyway, on which we make a compare later. But both `nonce` and `state` are required anyway. Personally for me it looks like a bug.

pavelmickevic:
Oops. I missed that part about tests. I will fix them for sure, a bit later.

ldesplat:
Well I've always wanted an OpenID Connect 1.0 implementation.

I'll need to review the full spec of OpenID Connect and make some decisions. If you need this right now, you'll have to maintain your own fork for a little bit.

lib/oauth.js
ldesplat: Well, it looks like you're implementing `3.3.  Authentication using the Hybrid Flow`
instead of the more traditional `3.1.  Authentication using the Authorization Code Flow` that bell assumes. I am talking specifically about the OpenID Connect spec @ http://openid.net/specs/openid-connect-core-1_0.html

In that case the nonce is indeed required. Also, not using a different value could allow an attacker to more easily identify the key we use to encrypt the cookie we keep around to validate the state. That's definitely a big problem.

PR #188, opened 2016-02-25
Add Auth0 provider

rolodato:
Wanting to use Auth0 with bell so I implemented it as a provider. If this could eventually get merged I can write the tests and add it to docs.

Some references:

* https://auth0.com/docs/protocols#oauth-server-side
* https://auth0.com/docs/auth-api
* https://auth0.com/docs/user-profile/normalized

Thanks!

### DIFF
--- a/lib/providers/auth0.js
+++ b/lib/providers/auth0.js
@@ -1,9 +1,16 @@
 'use strict';
 
+const Joi = require('joi');
+const Hoek = require('hoek');
+
 exports = module.exports = function (options) {
 
-    options = options || {};
-    const auth0BaseUrl = `https://${options.domain}`;
+    const validated = Joi.validate(options, Joi.object({
+        domain: Joi.string().hostname().required()
+    }));
+    Hoek.assert(!validated.error, validated.error);
+    const settings = validated.value;
+    const auth0BaseUrl = `https://${settings.domain}`;
     return {
         protocol: 'oauth2',
         useParamsAuth: true,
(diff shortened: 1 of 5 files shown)


lib/providers/auth0.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+exports = module.exports = function (options) {
+
+    options = options || {};
```
ldesplat: You should be enforcing that `domain` exists and is a string. See phabricator.js for an example.

ldesplat:
Add the docs and tests as well and I'll be very happy to accept this.

Thank you.

rolodato:
@ldesplat Tests and docs added.

ldesplat:
Thanks again :)

rolodato:
No problem, thank you! Looking forward to the new release :+1:

rolodato:
Thanks @travi. As of today Auth0 does not support configuring custom domains for login pages, but that's in our backlog (currently only supported for private instances which have their own domains instead of Auth0 subdomains).

Don't want to pollute this PR further, so if you want to discuss anything Auth0-related feel free to email me or create a thread at https;//ask.auth0.com. Thanks!

rolodato:
For anyone stumbling upon this PR, I've created a Hapi + Auth0 sample here: https://github.com/auth0/node-auth0/tree/2a07a14de7add7188a2d997fe9f2af09173d912d/examples/hapi-regular-webapp

PR #194, opened 2016-03-04
Support returning raw resource payload. Closes #193

### LINKED ISSUE #193: client.resource() raw payload
Technically, the `internals.parse()` method in oauth.js is broken because it assumes JSON payload always starts with a `{` character. This is needed to hack around broken OAuth implementations that return the wrong payloads with or without proper content-encoding headers. But it will break some APIs. This fix allows turning off the parser. A better solution would be to make `internal.parse()` better to actually provide value (but I don't have the time for that). Either way, this is a useful option.

hueniverse:


hueniverse:
@ldesplat I'm going to merge and publish this because I'm blocked on a project. If you have any concerns about this please let me know.

ldesplat:
Thanks! Would have been nice to add documentation but I see that part of the OAuth1 API is not actually documented, so I guess you're off the hook :)

hueniverse:
It's my own little secret API.

PR #196, opened 2016-03-07
Bitbucket incorrect id and styling inconsistency.

pygeek:
Not sure why this wasn't picked up in the PR, but there are a couple issues.
1. It seems `profile.uuid` should be used instead of `profile.username` for `credentials.profile.id`.
2. The style in which this was written differs from the other providers we have.

I've gone ahead and fixed both issues in this PR.

### DIFF
--- a/Providers.md
+++ b/Providers.md
@@ -78,11 +78,12 @@ Specific fields may vary depending on the identity provider used. For more infor
 The default profile response will look like this:
 
 ```javascript
-credentials.profile = {};
-credentials.profile.id = profile.username;
-credentials.profile.username = profile.username;
-credentials.profile.displayName = profile.display_name;
-credentials.profile.raw = profile;
+credentials.profile = {
+    id: profile.uuid,
+    username: profile.username,
+    displayName: profile.display_name,
+    raw: profile
+};
 ```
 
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -16,3 +16,8 @@ config.json
 coverage.*
 lib-cov
 complexity.md
+
+#Ignore vim temporary and swap files
```
gergoerdosi: Add these to your global gitignore that is specific to your environment, not to the project specific one. See more information here:

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

ldesplat:
Just remove your .gitignore file changes from this PR and I can merge this in.

Looking at the docs, it looks like v1 of the user endpoint only returned `username` and no uuid so maybe this is why? Thanks I'll merge it as soon as you make that change.

ldesplat:
I forgot, please update the documentation in `Providers.md` which has the same "bug"

pygeek:
I reverted the changes to the `.gitignore` and updated `Providers.md`—nice catch.

ldesplat:
Thank you very much :) I'll make a release tomorrow.

PR #202, opened 2016-04-05
Office365 provider

mattboutet:
This doesn't yet pass all the tests - `npm test` returns with 

     ...

Edit: Should be all set, first experience with Lab, and it took me a bit to track down the problems, but glad I did.


mattboutet:
Should be all set, first experience with Lab, and it took me a bit to track down the problems, but glad I did.

ldesplat:
That looks good! Thanks. Will merge it and make a release this week.

PR #213, opened 2016-05-09
Add Pinterest provider

acavestro:
Hi, 
I'm working on a project that needs to support Pinterest for authentication.

Since my team is using Bell to manage other social networks, I thought it would be nice to write a provider for Pinterest on my own and than submit it to you, so other people can use it.

I also wrote the test (100% coverage) and the documentation about it.

If you would like any further information, please don't hesitate to contact me.

Bye
Antonio

### DIFF
--- a/API.md
+++ b/API.md
@@ -65,7 +65,7 @@ server.register(require('bell'), function (err) {
 ### Options
 
 The `server.auth.strategy()` method requires the following strategy options:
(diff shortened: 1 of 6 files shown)


ldesplat:
Oh very cool! Big fan of Pinterest integration. Thank you so much!

PR #216, opened 2016-05-15
Consistently return node req object in oauth v1 client. Closes #215

### LINKED ISSUE #215: Fails to always return the node req object from wreck

hueniverse:
This might benefit from some tests confirming that all these methods return a node `req` object. This is needed in case you want to `req.abort()` etc. I don't have the time to add tests but this is probably worth merging and adding a new issue for the missing tests...


ldesplat:
I can't believe we don't even allow users to specify a timeout in request options or even send special headers (for tracking purposes) and such. Definitely something to improve upon.

Thanks! :)

PR #221, opened 2016-05-20
getMethod option

tameraydin:
Introduces ``getMethod`` option to Twitter provider in order to allow using different methods from its rest API.

### DIFF
--- a/lib/providers/twitter.js
+++ b/lib/providers/twitter.js
@@ -1,10 +1,14 @@
 'use strict';
 
+const Hoek = require('hoek');
+
 exports = module.exports = function (options) {
 
-    options = options || {};
-
-    const getMethod = options.getMethod || 'users/show';
+    const defaults = {
+        extendedProfile: true,
+        getMethod: 'users/show'
+    };
+    const settings = Hoek.applyToDefaults(defaults, options || {});
 
     return {
         protocol: 'oauth',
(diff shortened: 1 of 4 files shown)


lib/providers/twitter.js
```diff
@@ -4,6 +4,8 @@ exports = module.exports = function (options) {
 
     options = options || {};
```
ldesplat: Let's clean this up a bit now that it's getting slightly more complex
```
const defaults = {
    extendedProfile: true,
    getMethod: 'users/show'
};
const settings = Hoek.applyToDefaults(defaults, options);
```

Don't forget to require hoek and change options to settings.

lib/providers/twitter.js
```diff
@@ -20,7 +22,7 @@ exports = module.exports = function (options) {
                 return callback();
             }
 
-            get('https://api.twitter.com/1.1/users/show.json', { user_id: params.user_id }, (profile) => {
+            get('https://api.twitter.com/1.1/' + getMethod + '.json', { user_id: params.user_id }, (profile) => {
```
ldesplat: ES6 String
```
get(`https://api.twitter.com/1.1/${settings.getMethod}.json`, ...
```

ldesplat:
Looks good to me, Just these two minor comments. Thanks, now it makes more sense what you were asking for.

gergoerdosi:
I think the name `getMethod` is a bit confusing. Method has a meaning in requests and this setting has nothing to do with that.

tameraydin:
@ldesplat thanks for the feedback.. I did the changes you suggested but now somehow I'm getting this weird "'equal' of undefined" error on CI (when I try locally -node v4,5 or 6-, all tests are passing, no fail). Do you have any idea?

gergoerdosi:
@tameraydin: Remove `deep` in tests, it's the default in code v3.0.0.

tameraydin:
@gergoerdosi thanks, it helped! ..but I'm curious, why the assertion I added was the only one failing, I see that deep comparison is used in all other provider specs & assertions?

gergoerdosi:
I don't see `deep` being used anywhere in the code. Which files are you referring to?

tameraydin:
ahh my bad.. my fork branch wasn't up-to-date:) thx!

ldesplat:
Thank you @gergoerdosi for the consistent review. And thank you @tameraydin for contributing!

PR #208, opened 2016-04-23
Fix Safari CORS with Facebook & potentially other providers

ldesplat:
Based on #206 and Fixes #191.

Could I get someone to review this?

The crux of the issue is that Safari, due to CORS and the way Facebook sometimes does the redirect, it will not send the state cookie along. So, when we see no token (usually happens during development due to bad settings), we attempt to refresh the page using the meta keyword (not using javascript) and this ensures that Safari sends us the cookie.

We attempt to not have infinite loops with adding the refresh query parameter.

PR #225, opened 2016-06-01
Fix markup to quote URL correctly to avoid infinite redirect

akshayp:
@ldesplat After the fix for https://github.com/hapijs/bell/pull/208 it looks like at least on chrome it's possible to simulate an infinite redirect. It's triggered by unbalanced quotes used in the markup which make the current browser location be the redirect url without the refresh param. I've updated the markup to be similar to the [W3 Example](https://www.w3.org/TR/WCAG20-TECHS/H76.html#H76-examples) to avoid the infinite redirect


ldesplat:
Thank you for finding this. What a bad bug, from such a simple mistake.

akshayp:
No worries! Thank you for the quick review + merge

PR #227, opened 2016-06-03
Implemented RSA-SHA1 signatures for OAuth v1

### LINKED ISSUE #226: OAuth v1: How to handle RSA-SHA1 signing?
Hi!
I'm trying to implement JIRA authentication through bell. [JIRA](https://developer.atlassian.com/jiradev/jira-apis/jira-rest-apis/jira-rest-api-tutorials/jira-rest-api-example-oauth-authentication) uses RSA-SHA1 to create signatures. Unfortunately, it looks like bell uses HMAC-SHA1 [every single time](https://github.com/hapijs/bell/blob/master/lib/oauth.js#L457-L510), without letting us change the algorithm; is that correct?

RSA-SHA1 [is supported by `node-oauth`](https://github.com/ciaranj/node-oauth/pull/121), which `bell` seems to use.

Is there any way to change the signing algorithm? Thanks!

PaulMougel:
Here's a first draft which should fix #226. Feedback is welcome 👍


lib/index.js
```diff
@@ -52,6 +52,7 @@ internals.schema = Joi.object({
     password: Joi.string().required(),
     clientId: Joi.string().required(),
     clientSecret: Joi.string().required().allow(''),
+    signatureMethod: Joi.string().valid('HMAC-SHA1', 'RSA-SHA1').when('provider.protocol', { is: 'oauth', otherwise: Joi.forbidden() }),
```
ldesplat: I don't think there is a reason for a user of a provider would need to choose this option? If there is not, then it should be included in the provider object just above.

Also, when the protocol is oauth, do `then: Joi.default('HMAC-SHA1')` . This will ensure backwards compatibility and ensure it's always set. I know you take care of it  on line 382 of lib/oauth.js but why not let Joi handle it for you.

lib/oauth.js
```diff
@@ -378,7 +379,7 @@ internals.Client.prototype._request = function (method, uri, params, oauth, opti
     oauth.oauth_nonce = Cryptiles.randomString(internals.nonceLength);
     oauth.oauth_timestamp = Math.floor(Date.now() / 1000).toString();
     oauth.oauth_consumer_key = this.settings.clientId;
-    oauth.oauth_signature_method = 'HMAC-SHA1';
+    oauth.oauth_signature_method = this.settings.signatureMethod || 'HMAC-SHA1';
```
ldesplat: If you change the Joi validation, then this is not needed anymore and you can just set `this.settings.signatureMethod`

lib/oauth.js
```diff
@@ -503,10 +504,17 @@ internals.Client.prototype.signature = function (method, baseUri, params, oauth,
...
 
-    // HMAC-SHA1 (3.4.2)
-
-    const key = tokenSecret ? (this.settings.clientSecret + internals.encode(tokenSecret)) : this.settings.clientSecret;
-    return Crypto.createHmac('sha1', key).update(baseString).digest('base64');
+    let sig;
+    if (oauth.oauth_signature_method === 'RSA-SHA1') { // RSA-SHA1 (3.4.3)
+        const sign = Crypto.createSign('sha1');
+        sign.update(baseString);
+        sig = sign.sign(this.settings.clientSecret).toString('base64');
```
ldesplat: 1. Instead of doing `.toString` you can pass the `base64` to the 2nd parameter of sign.
2. I would also suggest you chain the calls to match the code below.
3. Format this slightly differently:
- instead of doing the else, just return from the function early. Then you don't need to create the `let sig` variable and I find we usually try to return early rather than doing if/else stmts.

test/index.js
```diff
@@ -117,6 +119,59 @@ describe('Bell', () => {
...
+                });
+
+                server.inject('/login?next=%2Fhome', (res) => {
+
+                    const cookie = res.headers['set-cookie'][0].split(';')[0] + ';';
+                    expect(res.headers.location).to.equal(mock.uri + '/auth?oauth_token=1');
+
+                    mock.server.inject(res.headers.location, (mockRes) => {
+
+                        expect(mockRes.headers.location).to.equal('http://localhost:80/login?oauth_token=1&oauth_verifier=123');
```
ldesplat: Due to the 100% test coverage and the other test cases, I know we exercise the `RSA-SHA1` path here, but I don't remember if there's a way to verify in the test case itself. ie. check the signature is RSA-SHA1? That will be useful long term and make this a slightly better test.

ldesplat:
Thank you so much. This looks great! I did an initial look over and made some comments. I've noted down a couple of things to check up on my side as well.

Now I just hope nobody asks for for the PLAINTEXT signature. Not one I am really enthusiastic about :)

PaulMougel:
Thank you for the comments.

Good call on moving the `signatureMethod ` to the provider configuration object. However, I don't see how I can do that without messing up with the `Mock` object, see `mock.js` on line 141:

```js
return callback({
    protocol: 'oauth',
    temporary: this.server.info.uri + '/temporary',
    auth: this.server.info.uri + '/auth',
    token: this.server.info.uri + '/token'
});
```

How can I send the new parameter without impacting every single test case?

ldesplat:
Apologies for the delayed response.

I think this can easily be done by adding a parameter / option to the Mock constructor. Then simply make the default what it currently is, and only when that parameter is supplied do you override. So then, you only have your testcases impacted.

PaulMougel:
Here you go! Tell me if that's better.

ldesplat:
Thank you very much. This looks perfect! :+1:

PR #229, opened 2016-06-08
Allow use of non-Google+ profile endpoints. (fixes #228)

### LINKED ISSUE #228: Remove dependency on Google Plus APIs
I was poking around to find out of there is a way to remove the dependency on Google Plus APIs now that Google is splitting Plus out of it's products. Reading Google's latest Identity Platform documentation, it seems that most endpoints are no longer associated with Google Plus at all.

https://developers.google.com/identity/protocols/OAuth2WebServer#callinganapi

While you can use a Google Identity auth key to call Google Plus APIs, such as requesting a users profile, Google Plus is not required for initial authentication. Additionally, the core APIs provide a way of getting a users basic information such as email address and given/family names.

Unfortunately, it appears that last year Bell moved in exactly the opposite direction #108. I suppose that's a side effect of Google making a big bet on Plus which didn't pan out in the end.

I believe the change required is just to modify the URL used to fetch the user profile https://github.com/hapijs/bell/blob/master/lib/providers/google.js#L13 and the code that parses out piece of the user profile (for example changing backward from `profile.emails` to `profile.email`.

There are two reason I'm interested in doing this:

1. I'm unsure of what Google Plus profiles will contain going forward, are new users automatically given a populated Google Plus profile? If not, will their gmail/docs/whatever details be provided instead or nothing at all?
2. Currently in order to use Bell, Google Plus APIs must be enabled on the developers account. This is an extra requirement over top of having an account and enabling an oAuth application and it's not actually required to get the details that users commonly retrieve for authentication.

Before making such a PR, I wanted to post this issue just to gauge interest. Perhaps I'm wrong in assuming this is a desirable change? Some 3rd party auth systems implement an additional options parameter for Google for the userProfileUrl to use and implement two different parsers depending on the URL provided. Would Bill rather have that as a feature, taking into consideration that that's one more feature to maintain down the road?

wwalser:
Fix for #228, allows developers to use profile URLs that don't depend on Google+ APIs being enabled.

See #228 for more details.

I implemented this change as an optional parameter in order to ensure it didn't change the default behavior of the Google provider. This ensures that the change can land without requiring a major version bump.

### DIFF
--- a/Providers.md
+++ b/Providers.md
@@ -193,10 +193,13 @@ The default profile response will look like this:
 
 [Provider Documentation](https://developers.google.com/identity/protocols/OpenIDConnect)
 
-You must also enable the Google+ API in your profile. Go to APIs & Auth, then APIs and under Social APIs click Google+ API and enable it.
+By default the Google provider uses a Google+ API to access a users profile. Using Google+ APIs requires developers to enable the Google+ API in the Google Developer Console. Go to APIs & Auth, then APIs and under Social APIs click Google+ API and enable it.
+
+Alternatively you may use the `profileUrl` config to fetch a users standard Google Account details. At the time of writing, [https://www.googleapis.com/oauth2/v3/userinfo](https://www.googleapis.com/oauth2/v3/userinfo) is the latest version of the alternative user profile endpoint.
 
 - `scope`: Defaults to `['profile', 'email']`
-- `config`: not applicable
+- `config`:
+  - `profileUrl`: String *optional* alternative to the default Google+ API profile URL. (use this if you wish to fetch non-Google+ profile data.)
 - `auth`: https://accounts.google.com/o/oauth2/v2/auth
 - `token`: https://www.googleapis.com/oauth2/v4/token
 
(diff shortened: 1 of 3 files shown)


ldesplat:
:+1: That's something I've also wondered about. Will take a closer look at the PR in the next few days.

ldesplat:
@qualquervalor You made https://github.com/hapijs/bell/pull/108 . Do you have any insights on this? Should we make another breaking change and make this the default again without Google+?

@wwalser I'll hold on to this for a little bit longer. It should not be too impactful in the moment since you can make your own provider but it appears we might be better off making a breaking change here. Maybe even support two google providers (Google & Google+) since the profile responses are a little bit/quite different.

wwalser:
This current change set is very minimal. It doesn't change the existing behavior, only adds a new option in case others want to opt out of G+. Would you like to me to create another PR with the API change to have two providers?

Also, I wasn't aware that one could create their own providers. Cool.

ldesplat:
Let's go ahead and have two providers. I prefer it over having yet another option that nobody knows about, especially one that requires you knowing the full url.

I'll close this PR and await your new changes. Thanks a lot!

Issue #230, opened 2016-06-10
Only publish root and lib directory in npm


ldesplat:
https://github.com/hapijs/contrib/issues/90

PR #231, opened 2016-06-10
Only publish root and lib directory in npm. #230

wwalser:
#230 
Adds an `.npmignore` file, uses @Marsup's [recommended](https://github.com/hapijs/contrib/issues/90) content since Bell has multiple files under `lib/`.


ldesplat:
thanks

PR #233, opened 2016-06-15
Allow providerParams to be a method invoked with the request

connor4312:
This allows provideParams to be a method which, when invoked with the request object, returns query parameters to add to the request.

For certain OAuth servers which include additional metadata in the request (for us: Discord's OAuth system) it's sometimes desirable to be able to selectively request features depending on configuration in the auth request.

### DIFF
--- a/API.md
+++ b/API.md
@@ -109,8 +109,7 @@ Each strategy accepts the following optional settings:
 - `isHttpOnly` - sets the cookie HTTP only flag. Defaults to `true`.
 - `ttl` - cookie time-to-live in milliseconds. Defaults to `null` (session time-life - cookies are deleted when the browser is closed).
 - `domain` - the domain scope. Defaults to `null` (no domain).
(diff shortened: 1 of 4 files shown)


ldesplat:
You're missing the documentation for the scope parameter.

Also, under Advanced Usage, would you mind adding a section describing/showing an example on how to do that for scope and providerParams.

Otherwise, it looks good to me. I was debating if we should try/catch around the user code... We could show a nice error message and throw back the original exception but I am not 100% positive that's worth it. If anybody has an opinion that it should, do speak up! :)

ldesplat:
Hi @connor4312 This is a great contribution. Could you please document the scope parameter changes?

connor4312:
Sorry for the delay, I've updated the documentation and added and example as request. Let me know if there's anything else to be tweaked 😄

ldesplat:
:+1: for the example

Very nice!

PR #235, opened 2016-06-16
Separate Google into two providers. #228

wwalser:
Issue is at: #228
See #229 for previous discussion.

### DIFF
--- a/Providers.md
+++ b/Providers.md
@@ -212,6 +212,30 @@ The default profile response will look like this:
 
 [Provider Documentation](https://developers.google.com/identity/protocols/OpenIDConnect)
 
+- `scope`: Defaults to `['profile', 'email']`
+- `config`: not applicable
+- `auth`: https://accounts.google.com/o/oauth2/v2/auth
+- `token`: https://www.googleapis.com/oauth2/v4/token
+
+The default profile response will look like this:
+
+```javascript
+credentials.profile = {
+    id: profile.id,
+    displayName: profile.name
+    name: {
+        given_name: profile.given_name,
+        family_name: profile.family_name
+    },
+    email: profile.email,
+    raw: profile
+};
+```
+
+### Google Plus
+
+[Provider Documentation](https://developers.google.com/identity/protocols/OpenIDConnect)
+
 You must also enable the Google+ API in your profile. Go to APIs & Auth, then APIs and under Social APIs click Google+ API and enable it.
 
 - `scope`: Defaults to `['profile', 'email']`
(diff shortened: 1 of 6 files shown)


Providers.md
```diff
@@ -193,6 +193,30 @@ The default profile response will look like this:
...
+
+The default profile response will look like this:
+
+```javascript
+credentials.profile = {
+    id: profile.id,
+    displayName: profile.name
+    name: {
+        given_name: profile.given_name,
+        family_name: profile.family_name
```
qualquervalor: These should be snake case `givenName` and `familyName`

lib/providers/google.js
```diff
@@ -10,13 +10,16 @@ exports = module.exports = function (options) {
...
 
                 credentials.profile = {
                     id: profile.id,
-                    displayName: profile.displayName,
-                    name: profile.name,
-                    emails: profile.emails,
+                    displayName: profile.name,
+                    name: {
+                        given_name: profile.given_name,
+                        family_name: profile.family_name
```
qualquervalor: These should be snake case `givenName` and `familyName`

wwalser: I don't have a problem with following that convention if that's what @ldesplat wants, but that will make this object different to Google Plus's api, which uses underscores. 

This is important because anyone upgrading from the previous version of Bell to this one will have to change their code to be snake case as well. Choosing code convention over API compatibility seems like the wrong call IMO.

test/providers/google.js
```diff
@@ -20,7 +20,7 @@ const expect = Code.expect;
 
 describe('google', () => {
 
-    it('authenticates with mock', { parallel: false }, (done) => {
+    it('Uses custom profileUrl', { parallel: false }, (done) => {
```
ldesplat: I guess this needs to be the old description

lib/providers/google.js
ldesplat: We unfortunately don't have proper conventions enforced through all the providers so I don't mind having it yet again different if it means a bit less breakage.

test/providers/google.js
wwalser: fixed

ldesplat:
Could you rebase this? I am sorry I merged another PR before this one. I'll be ready to make a release as soon as you do!

Thank you for waiting so patiently.

wwalser:
I'm out right now. Will get on this when home.

wwalser:
@ldesplat Rebased.

ldesplat:
Thank you so much!

PR #240, opened 2016-07-13
Fitbit provider

shellyb:
Fitbit uses Bell in their third-party integrations with Strava and IFTTT, and so we thought this would be a useful resource. Feel free to reach out if you have any questions or changes :) Thank you!


ldesplat:
Thank you, that looks great!

PR #244, opened 2016-07-18
New dropbox-v2 provider which uses Dropbox V2 API

klaronix:
The existing Dropbox provider uses Dropbox API V1 which has been deprecated in April 2016. New API V2 uses "RPC" style calls (i.e. POST) for requesting user account information. So far all providers are requesting account information by GET calls. There is a new provider config option now ("providerMethod") which can be set to 'post' in order to get account information by POST request (defaults to 'get').

[https://www.dropbox.com/developers/documentation/http/documentation#users-get_current_account](https://www.dropbox.com/developers/documentation/http/documentation#users-get_current_account)

### DIFF
--- a/API.md
+++ b/API.md
@@ -83,6 +83,7 @@ The `server.auth.strategy()` method requires the following strategy options:
       Defaults to space (Facebook and GitHub default to comma).
     - `headers` - a headers object with additional headers required by the provider (e.g. GitHub required the 'User-Agent' header which is
       set by default).
+    - `profileMethod` - `get` or `post` for obtaining user profile by `profile` function. Default is `get`.
     - `profile` - a function used to obtain user profile information and normalize it. The function signature is
       `function(credentials, params, get, callback)` where:
         - `credentials` - the credentials object. Change the object directly within the function (profile information is typically stored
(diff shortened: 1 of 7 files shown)


ldesplat:
This looks good on the first go. Maybe since you clearly pointed out that our existing dropbox implementation is no longer the preferred way, we can change this rather than adding a new dropbox-v2.

I depend on your to let us know if it breaks existing clients to do so, so that I know if this will be a major release or not.

What would make it a breaking release is if the profile response looks any different than before for example or if it supports different parameters than before. I would edge towards saying that yes it breaks existing clients and there is no issue for us to make a major release.

klaronix:
Unfortunately Dropbox V2 api breaks clients because of different profile response. So I would go for a '-v2' provider rather than urging bell users to migrate on next update. The old api will be switched off on 06/28/2017; maybe until then we should keep the two providers.

ldesplat:
I'll merge this in 8 days but if you could make the following modifications (i'll do them otherwise).

Use `Wreck.request` and pass the method instead of your current way. As well as default to `'get'` internally (along with the current default in the Joi schema).

Also, some documentation on the v2 provider would be appreciated.

ldesplat:
I have not forgotten about this PR. I've got some work in progress that will incorporate it :)

klaronix:
@ldesplat: Do you think, we get this PR integrated? Because I have another one ;=) It's commit 		36bd1f1 from today, which adds support to pass a custom Wreck agent (this allows to authenticate behind a proxy).

ldesplat:
Can you remove the custom agent from that branch and make another PR. I'll merge this when that's done.

klaronix:
OK. PR is now "agent free".

PR #249, opened 2016-08-05
Twitter additional GET parameters on extended profile request

bbatliner:
I was trying to retrieve the email of Twitter users and found it on [this API route](https://dev.twitter.com/rest/reference/get/account/verify_credentials). However, fetching the email from this route requires an additional GET parameter. This PR allows for additional parameters to that extended profile call.

Questions I have:

1. When writing the test for this option, I couldn't find a way in the Mock'd Twitter endpoint to verify if there were actual GET parameters passed to that route. Is this something that's testable?

2. Should dynamic GET parameters (via a function in the config) be allowed? In my use case, I only needed to add `"include_email": "true"`, but other parameters might be dependent on the authenticated user's information.

Thanks!

### DIFF
--- a/Providers.md
+++ b/Providers.md
@@ -466,6 +466,7 @@ credentials.profile = {
 - `config`:
   - `extendedProfile`: Request for more profile information
   - `getMethod`: [Twitter API](https://dev.twitter.com/rest/public) GET method to call when `extendedProfile` is enabled. Defaults to `'users/show'`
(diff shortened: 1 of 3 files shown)


ldesplat:
Thank you, this looks really useful. For dynamic parameters, nobody has asked for it and if they really want to, they can make their own profile call or submit a PR.

PR #252, opened 2016-08-15
[Update] Updates Facebook provider to v2.7

kamronbatman:
Updates Facebook provider to v2.7.
Tested with public profile options and is working.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,3 @@
 *
 !lib/**
+!.npmignore
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,18 @@ language: node_js
 node_js:
     - "4"
     - "6"
+    - "7"
     - "node"
 
 sudo: false
+
+env:
+    - HAPI_VERSION="13"
+    - HAPI_VERSION="14"
+    - HAPI_VERSION="15"
+    - HAPI_VERSION="16"
+
+install:
+    - npm install
+    - npm install hapi@$HAPI_VERSION
+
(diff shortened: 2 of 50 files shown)


ldesplat:
For reference this is the changelog between the APIs.. In this case going from 2.3 (supported until July 2017) to 2.7.

https://developers.facebook.com/docs/apps/changelog

Take a look at the changes in permissions, would explain the changes. In our case, it doesn't impact us but it does impact if people are passing in extra permissions which leads me to think that this is a breaking change for our users.

I'll merge this in a major release but I am going on vacation for 8 days so it will have to wait that much extra time for now!

kamronbatman:
Going from v2.3 to v2.7 is a big change.  There are a few deprecations and permission changes but that is minor compared to the improved performance and new endpoints/APIs.

Should we allow the user to specify the version through a field?  The Facebook provider is fairly simple and I think it can already support 2.x.

kamronbatman:
@hueniverse I noticed you removed this from the 9.0 milestone.  Any particular reason? Will it be added to a milestone after that?

Thanks!

ldesplat:
It will get done :) He removed it because he was doing a release of something else.

kamronbatman:
I realized that and that is why I removed my comment.  I updated it to the latest master.  I figure we can rebase or squash when ready.  Also since facebook v2.8 is out, this _might_ be outdated and need to be replaced with a new PR that has new testing.

kamronbatman:
Now that Facebook has moved beyond v2.7, should we consider a newer API version for the next milestone?

kamronbatman:
Closing this out in favor of #311.

PR #254, opened 2016-08-21
fix mapping of google profile fixes #251

### LINKED ISSUE #251: Google user info
Recently the google provider changed to use the oauth user info instead of google plus user info.
But as far as I can tell from https://developers.google.com/identity/protocols/OpenIDConnect#obtainuserinfo the id is now mapped to the `sub` property but it was not changed in the PR see https://github.com/hapijs/bell/commit/19e37eb5ba9ed50d359453b8910420f708b62459#diff-6cf5a5271f13794eed67d28ec16ffaa9L16

Am I doing something wrong or do you want a PR to fix this?

AdriVanHoudt:
This fixes #251 due of changes made in #235

### DIFF
--- a/lib/providers/google.js
+++ b/lib/providers/google.js
@@ -13,7 +13,7 @@ exports = module.exports = function (options) {
             get('https://www.googleapis.com/oauth2/v3/userinfo', null, (profile) => {
 
                 credentials.profile = {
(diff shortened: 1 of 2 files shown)


ldesplat:
Thanks, it's now in the 8.1.1 release.

PR #256, opened 2016-08-26
allow for leading whitespace in JSON

wraithgar:
Some oauth providers (specifically AT&T) return their JSON payloads with lots of leading whitespace, which fails the current test of `payload[0] === '{'` and causes the response from `get` to be passed through `Querystring.parse` instead.

This PR is an attempt to allow for situations like this, and still properly catch these payloads as JSON.

I tried looking through the tests to see where I could add a case to ensure this works but didn't see anywhere that stood out to me. I'm more than happy to attempt that if given a little more direction on this.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -558,7 +558,7 @@ internals.encode = function (string) {
 internals.parse = function (payload) {
 
     payload = Buffer.isBuffer(payload) ? payload.toString() : payload;
-    if (payload[0] === '{') {
+    if (payload.trim()[0] === '{') {
         try {
             return JSON.parse(payload);
         }


lib/oauth.js
```diff
@@ -558,7 +558,7 @@ internals.encode = function (string) {
 internals.parse = function (payload) {
 
     payload = Buffer.isBuffer(payload) ? payload.toString() : payload;
-    if (payload[0] === '{') {
+    if (/^\s*{/.test(payload)) {
```
nlf: just curious, but is calling `payload.trim()` to remove leading and trailing spaces enough to avoid the regex here?

can anyone think of any edge cases where maintaining leading or trailing whitespace would be necessary?

wraithgar: You could technically be clobbering Querystring params and/or values with whitespace if the test failed and it fell through to `Querystring.parse`.  Obviously that's super edge case and I've never personally seen query params with whitespace in their names, but that's anecdotal and this felt safer.

kamronbatman: Considering trim() is nondestructive to the original reference and JSON.parse() ignores leading/trailing whitespaces, then I don't see the possible clobbering.  Why wouldn't something like this work?

```
payload = Buffer.isBuffer(payload) ? payload.toString() : payload;
if (payload.trim()[0] === '{') {
```

wraithgar: Yes you're right. For some reason I was assuming nlf was talking about using it destructively a la `payload = payload.trim()`.  No clue why I thought that. I'll switch it to use trim instead.

kamronbatman:
@nlf I think this can be merged, what do you think?

nlf:
looks good to me

ldesplat:
Thank you so much! Yeah the JSON detection is definitely not the best.

PR #265, opened 2016-09-20
Set isSameSite to false and add code to support Hapi 13.5+

### LINKED ISSUE #264: Bell is not compatible with Hapi 15 due to same-site: strict for cookie
Bell is currently not compatible with Hapi 15. You should stick with Hapi 14 until we set the same-site attribute for the cookie that Bell has to set to either be turned off or set to lax. I am experimenting with different providers to determine how they do the redirect back, but until then hapi 15 will not work.

ldesplat:
Resolves #264 in a non-breaking manner.

Had to turn off code coverage for the Hapi 13 support, due to multiple reasons:
1. Don't want to run code coverage using hapi 13
2. Hapi 14 and up, don't error on server.state with cookie passwords less than 32 characters like Hapi 13 did so I don't know how to force an error here either.
3. The way Hapi is implemented makes it hard to stub out the server.state method to force those scenarios.

Plan is to remove this code in next breaking release and stop support for Hapi 13 at that time.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -105,6 +105,7 @@ internals.implementation = function (server, options) {
         password: settings.password,
         isSecure: settings.isSecure !== false,                  // Defaults to true
         isHttpOnly: settings.isHttpOnly !== false,              // Defaults to true
+        isSameSite: false,
         ttl: settings.ttl,
         domain: settings.domain,
         ignoreErrors: true,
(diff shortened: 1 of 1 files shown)


ldesplat:
Believe it or not, thanks to some great work from others, Strict did work with every browser except Chrome. It would fail the first time, but then we initiate a redirect to ourselves (only once) and all the browsers would properly send the Strict cookie :)

AdriVanHoudt:
So it would fail on every browser but then you do a redirect to yourself and it works? I assume this is because then you become the initiator and then it is ok to send the strict cookie along since it is now the same domain. But Chrome did not do this?

ldesplat:
Correct. Chrome just refused to send the cookie when we became the initiator for Strict.
For Lax it should just work from the beginning and it did with all browsers but Chrome again! Could not work on my machine.

AdriVanHoudt:
I just tried to put it to Lax and it does seem to work on Chrome Version 53.0.2785.116 (64-bit) on macOS,
changing this would be at least a minor (if not major) since this can have some implications. Could you test again maybe it was a issue in your version of chrome?

PR #273, opened 2016-10-13
Fixes #261 expose OAUTH2 token response as request.auth.artifacts

### LINKED ISSUE #261: Support for reading id_token from an OpenId Connect flow
Hey.

[OpenId Connect](http://openid.net/connect/) starts out as an oAuth2 flow and bell can therefor be reused for the initial phase. (authentication code flow)

One difference is that OpenID connect can exposes two tokens, one access token but also an id token. It would be fantastic if there was a way to read this id token.

I'm not suggesting to add full OpenID Connect support but I would be able to use the bell library if the payload from the token endpoint was exposed, for example, as an artifact.

More then happy to put together a PR if this sounds fair / reasonable.

garthk:
OpenId Connect providers return an ID token in the `id_token` property of
their `token` response, e.g. `auth0` with `scope: ['openid']`. This
change caters for any scenario requiring unexpected properties from the
`token` response by exposing the full `token` payload as
`request.auth.artifacts`.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -262,7 +262,7 @@ exports.v2 = function (settings) {
             };
 
             if (!settings.provider.profile || settings.skipProfile) {
-                return reply.continue({ credentials });
+                return reply.continue({ credentials, artifacts: payload });
             }
 
             // Obtain user profile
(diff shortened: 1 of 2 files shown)


ldesplat:
That looks good to me on first view. Let me play with it this week.

garthk:
Rebased.

jsatk:
I'd love to see this get merged. We would need something like this for our uses. I opened a PR to add Okta and referenced this PR in it. https://github.com/hapijs/bell/pull/299

PR #284, opened 2016-12-19
+ weibo and qq providers

evilRoma:
added QQ and Weibo providers


ldesplat:
For some reason, I never commented on this. If you add documentation in the Providers.md file then we can add this. It's been many months so I'll close it but feel free to reopen. Entirely my fault for not reviewing it!

PR #289, opened 2017-01-11
AzureAD: Support email, when userPrincipalName (upn) is not available

taktran:
Support the email field, when `userPrincipalName` is not available.

From the Azure Active Directory graph API documentation about `userPrincipalName` (`upn`):

> This property is required when a work or school account is created; it is optional for local accounts. 

https://msdn.microsoft.com/en-us/library/azure/ad/graph/api/entity-and-complex-type-reference#contact-entity

When `upn` is not available, `profile.email` stores the email, so fall back on this value.

Have tested on an account created from live.com, where the `email` value exists there, but not `upn`.

### DIFF
--- a/Providers.md
+++ b/Providers.md
@@ -82,7 +82,7 @@ The default response would look like this in the `profile` object obtained
 credentials.profile = {
     id: profile.oid,
     displayName: profile.name,
-    email: profile.upn,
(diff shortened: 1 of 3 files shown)


ldesplat:
Thanks for the PR. It finally makes it in :)

PR #303, opened 2017-03-21
Facebook

doron2402:
- Using  Facebook newest api version (2.8)
- Adding user picture to request

### DIFF
--- a/lib/providers/facebook.js
+++ b/lib/providers/facebook.js
@@ -10,15 +10,15 @@ exports = module.exports = function (options) {
     return {
         protocol: 'oauth2',
         useParamsAuth: true,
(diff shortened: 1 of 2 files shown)


AdriVanHoudt:
is there anything breaking about the upgrade to 2.8? would existing implementations just keep working?

doron2402:
@AdriVanHoudt ,
not that I know of but I've updated that locally and it seems to be working fine.

kamronbatman:
First, we should look into v2.9, which doesn't look like it has anything breaking between v2.7 and v2.8.  Also this PR requires updates to oauth.js and tests.

ldesplat:
#311 got merged which takes care of this. I know I took a very long time, but thank you so much for persevering :)

PR #307, opened 2017-03-30
Fix Okta example

jsatk:
My example for Okta was incorrect. It was missing a required `config` with a `uri` key.


jsatk:
@ldesplat @AdriVanHoudt Please let me know if you have any feedback. Thanks!

ldesplat:
@jsatk I just modified it so it can pass linting by adding spaces in between the brackets. Thanks!

jsatk:
Thank you! Sorry about that!

PR #299, opened 2017-02-27
Add support for authentication with Okta

jsatk:
I need support for Okta with Bell. I've currently forked it and have patched it in, but I think this would be useful for the mainline bell release.

Our use of Okta also depends on exposing the payload. It appears that that was already done in this PR. https://github.com/hapijs/bell/pull/273

I'd love to see https://github.com/hapijs/bell/pull/273 land and then this. 

I tried my best to follow all current style and best practices in Bell. Please @ldesplat let me know if you have any requests, feedback, etc.

PR #308, opened 2017-04-03
Add `'profile'` to list of default scopes for Okta

jsatk:
The overwhelming majority of use cases for Okta will need profile information. It was an oversight on my part to not add this originally here: https://github.com/hapijs/bell/pull/299.


ldesplat:
Thanks!

PR #313, opened 2017-05-25
Update to Salesforce Provider

phitran:
UserInfo endpoint support for profile fetch. Provider was using the Force.com Identity Service prior.

https://developer.salesforce.com/page/Inside_OpenID_Connect_on_Force.com#User_Profile_Service


ldesplat:
This is a very nice improvement!

PR #314, opened 2017-05-31
Added mixer provider

jerome724:
Tested, seems like it is working. Would appreciate it when someone else could test it.

### DIFF
--- a/lib/providers/mixer.js
+++ b/lib/providers/mixer.js
@@ -1,0 +1,32 @@
+'use strict';
+
+exports = module.exports = function (options) {
+	options = options || {};
+	const uri = options.uri || 'https://mixer.com/';
+	const user = options.uri ? options.uri + 'api/v1/users/current' : 'https://mixer.com/api/v1/users/current';
+
+	return {
+		protocol: 'oauth2',
+		auth: uri + 'oauth/authorize',
+		token: uri + 'api/v1/oauth/token',
+		scope: ['user:details:self'],
+		useParamsAuth: true,
+		scopeSeparator: ',',
+		headers: {
+			'User-Agent': 'hapi-bell-mixer'
+		},
+		profile: ((credentials, params, get, callback) => {
+
+			get(user, null, (profile) => {
+				credentials.profile = {
+					id: profile.id,
+					username: profile.username,
+					email: profile.email,
+					raw: profile
+				};
+
+				return callback();
+			})
+		})
+	}
+};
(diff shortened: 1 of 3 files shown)


lib/providers/mixer.js
```diff
@@ -0,0 +1,32 @@
...
+	var uri = options.uri || 'https://mixer.com/';
+	var user = options.uri ? options.uri + 'api/v1/users/current' : 'https://mixer.com/api/v1/users/current';
+
+	return {
+		protocol: 'oauth2',
+		auth: uri + 'oauth/authorize',
+		token: uri + 'api/v1/oauth/token',
+		scope: ['user:details:self'],
+		useParamsAuth: true,
+		scopeSeparator: ',',
```
connor4312: Not used by default, but this should be a space. We follow [the spec](https://tools.ietf.org/html/rfc6749#section-3.3) property :)

jerome724: what is not used by default? and yes, just got a message in discord, sorry.

connor4312: The default scope is set to `user:details:self`, so there's not multiple scopes to separate. However if a consumer overrides this and requests additional scopes, they'll get rejected.

jerome724:
https://travis-ci.org/hapijs/bell/jobs/238105605
https://travis-ci.org/hapijs/bell/jobs/238105607

Is that an issue on my side?

yarn.lock
```diff
@@ -0,0 +1,1294 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
SimonSchick: Did you mean to commit this?

jerome724: oh no, that got generated when i installed the needed dependencies with yarn, thought that is already in the ``.gitignore``, removing it asap.

lib/providers/mixer.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+exports = module.exports = function (options) {
+
+    options = options || {};
+
+    const uri = options.uri || 'https://mixer.com';
+    const user = options.uri ? options.uri + '/api/v1/users/current' : 'https://mixer.com/api/v1/users/current';
```
sholladay: This should probably just be:

```js
const user = uri + '/api/v1/users/current';
```

ldesplat:
Thank you for this, it's now merged through #314

Issue #184, opened 2016-02-07
Simplify redirect_uri / location / protocol


ldesplat:
Proposal:
- Remove forceHttps option
- only `location` is now accepted to change the `redirect_uri`.

This simplifies the code greatly. Removes all guesses about protocol and lots of if checks.
- `location` can also be a function `function (request: Request) -> location: String`
  Allows people to specify a custom dynamic location string. The function will be called twice during an OAuth step. Therefore need to explain how to get the necessary information for dynamic locations from the request object. Prevents us from having to store the location information.

@nouellette @klaronix : Would this allow your use cases?

Alternatively the location function could only be called once and we internally store it in the state. It would only impact the code path when providing a location function but I don't see why if you can provide the right redirect_uri the first time, that you can't do it again.

Also, the function being synchronous is that ok? I don't want to do async because then have to involve domains to ensure we catch any thrown errors :(

nouellette:
If you could allow the plugin to accept a request.query.location to update that property similar this this check https://github.com/hapijs/bell/blob/master/lib/oauth.js#L161 but after it does the check it would call the update location function synchronously. That would work for my use case.

Thanks for looking into this further :)

klaronix:
Unfortunately this won't work for my case: I would like to provide a constant `redirect_uri` (of course, dynamic is even better) to a central "helper" server. Scenario is like this:

1. Bell protected hapi application server on "a.com" receives authorization request
2. "a.com" has a Bell location setting which sets `redirect_uri` to an endpoints on another server at "b.com"
3. "b.com" server determines "a.com" uri from state query parameter (I assume this is the only way since redirect is handled by browser and OAuth spec doesn't allow another "general purpose" parameter) and redirect back to "a.com" with all received params
4. "a.com" successfully completed authorization call

"b.com" doesn't need to run bell, not even hapi; it only has to know, how to extract location of "a.bom" from state parameter. I'd like to have this because of inflexible redirect configuration of certain OAuth service providers in combination with multiple backend application servers which are authorizing client requests.

ldesplat:
Bumping this to 8.0.0. If anybody wants to implement it or a version of it, they are free to do so. I am not going to.

dpmott:
I need part of this functionality; I'll make a PR to allow location to be a function.

PR #319, opened 2017-07-27
Allowing the location option to be a function

dpmott:
Addresses some of the need described in https://github.com/hapijs/bell/issues/184.

Specifically, I need this to avoid appending the request.path to the redirect URI, because I'm working with an OAuth2 provider that forbids any path (even a trailing slash).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -73,7 +73,10 @@ internals.schema = Joi.object({
     profileParams: Joi.object(),
     skipProfile: internals.flexBoolean.optional().default(false),
     forceHttps: internals.flexBoolean.optional().default(false),
-    location: Joi.string().optional().default(false),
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -73,7 +73,10 @@ internals.schema = Joi.object({
     profileParams: Joi.object(),
     skipProfile: internals.flexBoolean.optional().default(false),
     forceHttps: internals.flexBoolean.optional().default(false),
-    location: Joi.string().optional().default(false),
+    location: Joi.alternatives().try(
```
ldesplat: This does do the same thing as it did before in the case of not specifying a location. As in, without a location option, location will be undefined. I'll fix that since it's easy.

PR #335, opened 2017-11-29
refactor: cleanup syntax errors from async refactor

ChristianMurphy:
remove extra closing braces.
ensure `await` is always contained in an `async` function.


ldesplat:
That's awesome! we'll get there :D

PR #340, opened 2017-12-31
updated tests/implementations to work with hapi 17

AVVS:
Did some refactoring, all tests pass, though I'd clean up code a bit

### DIFF
--- a/examples/arcgisonline.js
+++ b/examples/arcgisonline.js
@@ -1,46 +1,39 @@
-'use strict';
-
 // Load modules
 
 const { Server } = require('hapi');
-const Hoek = require('hoek');
 const Bell = require('../');
 
+(async () => {
 
-const server = Server({ port: 8000 });
+    const server = Server({ port: 8000 });
 
-await server.register(Bell);
+    await server.register(Bell);
 
-Hoek.assert(!err, err);
-server.auth.strategy('arcgisonline', 'bell', {
-    provider: 'arcgisonline',
-    password: 'cookie_encryption_password_secure',
-    isSecure: false,
-    // You'll need to go to https://developers.arcgis.com/en/applications and set up an application to get started
-    // Once you create your app you will get your ClientID and Client Secret.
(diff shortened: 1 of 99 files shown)


metoikos:
I've tested with Twitter in a live app and worked like a charm.

examples/arcgisonline.js
```diff
@@ -1,46 +1,39 @@
...
-    path: '/bell/door',
-    config: {
-        auth: 'arcgisonline',
-        handler: function (request, h) {
-
-            reply('<pre>' + JSON.stringify(request.auth.credentials, null, 4) + '</pre>');
+    server.auth.strategy('arcgisonline', 'bell', {
+        provider: 'arcgisonline',
+        password: 'cookie_encryption_password_secure',
+        isSecure: false,
```
benatkin: This should be true for the purposes of an example IMHO, with a comment that it should only be set to false for development.

benatkin:
I got GitHub OAuth working with this. 👍🏻

examples/arcgisonline.js
AVVS: Well, I merely updated to async/await, and did some linting (ie top-level await is not yet possible, hence the wrap)

wraithgar:
I've got this working happy-path with quite a few providers, including a custom one.  However there is a change between this branch and the current bell behavior that I think is worth bringing up.

If the request that bell makes to the `temporary` url (and probably others, haven't tested those) has an error, the response code from that request will bubble up to the client making the request.

The behavior used to be (correctly I would think) that a 500 error would be sent back to the client.

marpstar:
FWIW, I tested this with LinkedIn tonight and it worked great.

qraynaud:
Working great with Google too. Thanks for the hard work!

flipxfx:
Any timeline on when this will be merged? Can't use hapi 17 without bell.

qraynaud:
I think that @wraithgar is correct about the fact that an error generated by a provider should result on an error 500 and not bubble up to the client.

What are your thoughts on this @AVVS ? And are you willing to fix this if you are in agreement over this ?

PR #350, opened 2018-03-06
Picture object included in the profile object for facebook oauth

satnam-sandhu:
Hi @AdriVanHoudt ,
Finally after changing the picture fields' value from just the url to a object i am able to pass the tests.
Sorry for bothering you again, but this time it seems to be working fine.
Thank you. :)


satnam-sandhu:
Yes it does match the test. I have modified the tests accordingly.

PR #358, opened 2018-04-11
Use OIDC compliant profile from Auth0

sholladay:
This tiny change ensures that Auth0 won't break for bell users when Auth0 replaces their legacy user profile endpoint with an OIDC compliant one.

Note that this change does not require any action on the part of the user and it is backwards compatible because Auth0 is already returning `profile.sub` in addition to the old `profile.user_id` in order to make it easier to transition.

### DIFF
--- a/lib/providers/auth0.js
+++ b/lib/providers/auth0.js
@@ -26,7 +26,7 @@ exports = module.exports = function (options) {
         token: `${auth0BaseUrl}/oauth/token`,
         profile: async function (credentials, params, get) {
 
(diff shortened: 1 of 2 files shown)


sholladay:
Done. :)

AdriVanHoudt:
Thank you!
Do you need this change quickly? Otherwise I will queue it up with other changes for the release.

sholladay:
Nope, no hurry. Auth0 is not forcing OIDC onto anyone for at least the next couple of months. This change is mostly to make `bell` more future proof, in light of the fact that the old way is officially deprecated.

AdriVanHoudt:
Awesome, thanks!

PR #363, opened 2018-05-23
[WIP] Usage without a strategy

optii:
This is a work in progress, to allow the usage of bell without a strategy. (see slack discussion [https://hapihour.slack.com/archives/C6ATVCFC0/p1526648765000150](here) with @AdriVanHoudt ) opening so we can discus this more

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -29,6 +29,16 @@ exports.v1 = function (settings) {
         const name = settings.name;
         const protocol = internals.getProtocol(request, settings);
 
+        // Replace h.authenticated and h.unauthenticated if not invoked via a strategy
+
+        if (!h.authenticated){
+            h.authenticated = (credentials) => Promise.resolve(credentials);
+        }
+
+        if (!h.unauthenticated) {
+            h.unauthenticated = (error, credentials) => Promise.reject({ credentials, ...error });
+        }
+
         // Prepare credentials
 
         const credentials = {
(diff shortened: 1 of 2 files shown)


AdriVanHoudt:
* Would it not be easier to just define the missing functions at the start of v1 or v2 instead of having to call the internal one everywhere? (just a thought)
* `error.reformat` doesn't add the credentials to the error, I think just doing `err.credentials = credentials` is fine.
* if `unauthenticated` rejects, `authenticated` should resolve a promise no?

hueniverse:
This is twisted. Why can't you just define a strategy and then use the hapi auth api to manually use it in the handler?

optii:
@AdriVanHoudt : 

 - Yes that could be an easier solution
 - ok didn't realize that for `error.reformat` will make the necessary changes
 - And yes that would make total sense :+1: 

i'll look at making the changes, if you still think this is something you want to see in bell

@hueniverse : 

The reason behind this is because the credentials of the provider are dynamic in my case  (multi-tenancy), so I would have to keep the strategies up to date whenever they got updated or created, i mean this is feasible, but i thank that adds unnecessary complexity.  could possibly also be missing a trick

hueniverse:
I don't want this change. I think you should find a way to wrap/achieve this from the outside.

optii:
Ok, in that case i'll close the PR.  Maybe the API docs could be updated as they make out that this is possible. if you'll accept a PR for that, i'll happily do one

AdriVanHoudt:
@hueniverse as in use https://hapijs.com/api#-await-serverauthteststrategy-request?
But then you need to create strategies on the fly I guess.
I think we should remove it from the docs then and maybe don't export the v1 and v2 function, since it doesn't make sense then anymore.

PR #368, opened 2018-06-07
Update bell.simulate() docs to align with hapi 17 updates

travi:
looks like this wasnt updated when the signature changed for hapi v17. i spent a bit more time figuring this out while upgrading an app to hapi v17 because the docs weren't up to date. since this change aligns well with other hapi v17 updates, i expected it to be similar to this, but had to dive into the source to confirm the expectation.

this was somewhat awkward to adjust and still stay in line with the conventions of the rest of the document, so i'm open to adjustments if there are thoughts that would make it more consistent.


AdriVanHoudt:
Looks good, thanks!

PR #372, opened 2018-07-04
V9.3.2 add config 'fields' to vk provider

slaawwa:
Update response data from vk (move 'uid' to 'id')
And add custom config 'fields'

### DIFF
--- a/lib/providers/vk.js
+++ b/lib/providers/vk.js
@@ -12,7 +12,7 @@ exports = module.exports = function (options) {
 
     options = options || {};
     const version = options.version || '5.73';
-    const key_id = parseInt(version) < 5? 'uid': 'id';
+    const key_id = parseInt(version) < 5 ? 'uid' : 'id';
 
     return {
         protocol: 'oauth2',
--- a/lib/providers/vk.js
+++ b/lib/providers/vk.js
@@ -26,11 +26,15 @@ exports = module.exports = function (options) {
                 v: version
             };
 
+            if (options.fields) {
+                query.fields = options.fields;
+            }
+
             const data = await get('https://api.vk.com/method/users.get', query);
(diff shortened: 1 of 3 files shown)


lib/providers/vk.js
```diff
@@ -26,11 +26,15 @@ exports = module.exports = function (options) {
...
+            if (options.fields) {
+                query.fields = options.fields;
+            }
+
             const data = await get('https://api.vk.com/method/users.get', query);
             const profile = data.response[0];
 
             credentials.profile = {
-                id: profile.uid,
+                id: profile.id,
```
AdriVanHoudt: Afaik the api returns uid not id?
https://vk.com/dev/fields

slaawwa: Page move from
https://vk.com/dev/fields
to
https://vk.com/dev/objects/user

```
https://vk.com/dev/users.get

Returns a list of user objects. 

A deactivated field may be returned with the value deleted or banned if a user has been suspended.
```
I think vk team move field  **uid** to field **id** in last version

AdriVanHoudt: I can get to both pages by clicking through the docs so confused haha
Do you use the api? Can you validate that this change works?

slaawwa: @AdriVanHoudt, yes. I use it) And this is reason why I create this pull-request)

slaawwa: @AdriVanHoudt, maybe try get `id` field and if it's field not exist try get `uid` field?

AdriVanHoudt: Sounds like a plan, if you change that I will merge 🎉

slaawwa: @AdriVanHoudt, what the problem with travis tests?

PR #373, opened 2018-07-08
Use Stripe Express

sholladay:
"Express", as Stripe calls it, is a new flow for setting up Stripe Connect accounts. It is more friendly to users and looks better. Stripe is recommending using this going forward, though the older "Standard" account setup is still available.

I'm not sure what the semver policy is here for changes to providers, but this should probably be considered a breaking change for Stripe users, to be safe, since there are some slight differences when interacting with Stripe Express accounts via their API. There are minimal changes for `bell` users, though.

### DIFF
--- a/lib/providers/stripe.js
+++ b/lib/providers/stripe.js
@@ -15,9 +15,9 @@ exports = module.exports = () => {
     return {
         protocol: 'oauth2',
         useParamsAuth: true,
-        auth: `https://${domain}/oauth/authorize`,
+        auth: `https://${domain}/express/oauth/authorize`,
         token: `https://${domain}/oauth/token`,
-        scope: ['read_only'],
+        scope: [],
         headers: { 'User-Agent': 'hapi-bell-stripe' },
         profile: async function (credentials, params, get) {
 
--- a/lib/providers/stripe.js
+++ b/lib/providers/stripe.js
@@ -10,11 +10,11 @@ const Joi = require('joi');
 const internals = {
     schema: Joi.object({
         express: Joi.boolean().optional()
-    }).required()
+    }).optional()
 };
 
 
-exports = module.exports = (options) => {
+exports = module.exports = (options = {}) => {
 
     const settings = Joi.attempt(options, internals.schema);
     const domain = 'connect.stripe.com';
--- a/lib/providers/stripe.js
+++ b/lib/providers/stripe.js
@@ -2,20 +2,28 @@
 
 // Load modules
 
+const Joi = require('joi');
+
 
 // Declare internals
 
-const internals = {};
+const internals = {
+    schema: Joi.object({
+        express: Joi.boolean().optional()
+    }).required()
+};
 
 
-exports = module.exports = () => {
+exports = module.exports = (options) => {
 
+    const settings = Joi.attempt(options, internals.schema);
     const domain = 'connect.stripe.com';
+    const authPrefix = settings.express ? '/express' : '';
 
     return {
         protocol: 'oauth2',
         useParamsAuth: true,
-        auth: `https://${domain}/express/oauth/authorize`,
+        auth: `https://${domain}${authPrefix}/oauth/authorize`,
         token: `https://${domain}/oauth/token`,
         scope: [],
         headers: { 'User-Agent': 'hapi-bell-stripe' },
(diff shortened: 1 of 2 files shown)


AdriVanHoudt:
I do love me some shiny new flows ✨
Do you have some docs or a changelog about this change?

sholladay:
Users will want to review these:
 - [Requirements for using Express](https://stripe.com/docs/connect/express-accounts#requirements-for-creating-express-accounts)
 - [Integrating with the Express Dashboard](https://stripe.com/docs/connect/express-dashboard)
 - [GitHub repo of the official Express demo](https://github.com/stripe/stripe-connect-rocketrides)

The extent to which the above affects an app depends on how Stripe is being used. Most apps will simply need to provide a link to the Express dashboard for their users.

AdriVanHoudt:
Hmm this kinda bothers me `Users based in the United States` :/
Also this would be a breaking change.
Would it be an option to add this as a separate provider?

sholladay:
Companies outside the US are able to contact support to have their account set up manually. Removing that step is also on the roadmap.

> this would be a breaking change

Agreed. Our app kept working as-is after this change, though.

AdriVanHoudt:
I'm guessing your app is US based then? :P
I know for a fact that there are companies using 1 year old Stripe api versions but maybe those are outliers.
I guess we can go with a major version and then document and point to those links.

hueniverse:
I would add this as a new provider for now, not replacing the existing one.

sholladay:
IMO, if the breaking change is too unpalatable, then a `useExpress` option is the way to go. Stripe support has made it clear to me that they are advising all apps to move away from "Standard" accounts. I'm not sure it's fair to call the old flow officially deprecated yet, but it's days are numbered. All new accounts should be using Express and thus I think `bell` should move to this, otherwise new users are setting themselves up for more pain later, possibly without realizing it.

hueniverse:
On its own, doesn't seem like this is worth a breaking change release. How you choose to do this is up to you.

sholladay:
Okay, this is ready for review again. I rebased and added an `express` option that maintains compatibility. This is no longer a breaking change. Hoping this can go away in a future release, though. :)

sholladay:
CI failed only for Node 11 and I don't think it's related to this PR. The output is:
> The following leaks were detected:queueMicrotask

hueniverse:
Need lab 17.

AdriVanHoudt:
I like the option but don't make it required to pass an object into Bell.providers.stripe() cause that would be breaking and doesn't have to be I think.

It is weird that Stripe is advocating to move to Express if it is only available for US entities :D /shrug

Rest LGTM

Nargonath:
Perhaps it would be nice to display a deprecation warning when not using the `express` option so people are aware of this change to come if not already.

AdriVanHoudt:
As long as this is on their website
```
While platforms located anywhere Stripe is available can use Express, the Express account itself can only be created for individuals, businesses, and non-profits based in the United States.
```
I don't think I want that. If Stripe is so persistent on switching they will contact their customers anyway no? I don't think that is Bell's responsibility.

Nargonath:
Yeah I think they'd contact their customers directly for sure but perhaps that'd be a good thing to already allow such behavior without replacing the old one. That way we don't break anything for the existing users and those that want to use `express` can do so. `express` will eventually replace the old way it is just a matter of time as I understood. Let's just get ahead of it.

AdriVanHoudt:
I'm totally on board with adding this. If you don't require the options object this is good to go for me.

sholladay:
I will make the object optional. Earlier, I noticed that all of the other providers that take an object require it. I just copied what the others were doing on the assumption that bell would pass an empty object if needed. But yeah, the provider can be used more directly, so I guess that makes sense despite the inconsistency.

sholladay:
Okay, I pushed the requested changes. Ready for review.

For future reference, Stripe will manually set you up if you need an Express account or customers outside of the United States. Their documentation is overly cautious so as to set expectations. It was one of our first concerns when signing up. They are working on automating it but in the meantime, it's a manual process and you have to contact them and ask.

sholladay:
Woah, I'm getting completely different results locally (where everything is passing) vs CI. Will do that Lab upgrade...

PR #374, opened 2018-07-16
Ensure Auth0 profile is availabe in all cases

sholladay:
This is a bugfix for Auth0’s “Last time you logged in with” flow. Previously, if a user clicked on their existing session instead of the "Not your account?" link, then `request.auth.credentials` would not contain the user's profile information. That was very confusing because `bell` normally provides the full profile. The bug was a hard one to track down, since it only happens to users who visit the login page while they are already logged in.

This PR fixes the problem by explicitly requesting all of the scopes necessary for `bell` to function as intended. Auth0's APIs behave themselves when you do this. 😄

### DIFF
--- a/lib/providers/auth0.js
+++ b/lib/providers/auth0.js
@@ -24,6 +24,7 @@ exports = module.exports = function (options) {
         useParamsAuth: true,
         auth: `${auth0BaseUrl}/authorize`,
         token: `${auth0BaseUrl}/oauth/token`,
+        scope: ['openid', 'profile', 'email'],
         profile: async function (credentials, params, get) {
 
             const profile = await get(`${auth0BaseUrl}/userinfo`);  // https://auth0.com/docs/api/authentication#user-profile


AdriVanHoudt:
So normally you don't need these scopes unless you come on the login page while logged in and you try to log in with the current session?
Seems like a bug on Auth0's end?

I don't know Auth0's api but if you say this is a no-op for other situations then this one I will gladly merge this :D

sholladay:
I have a slight suspicion that it has to do with OIDC conformance, which Auth0 has been slowly moving towards over the past year, but I can't easily prove that. I just remember it working prior to some of their OIDC changes. Also, their [scopes](https://auth0.com/docs/scopes/current) documentation seems to indicate these scopes are required to return the profile and email. But for some reason, in practice, it isn't necessary when creating a new session. Maybe someone made a mistake while trying to maintain backwards compatibility.

I will also note that the Okta provider uses these scopes, which seems relevant, since their product has a similar architecture.

AdriVanHoudt:
Sounds good 👌 
I asked in Slack for some Auth0 peeps to look at this but I think you are right. I'm waiting on some other PR's so I can do a release with some changes combined.
I expect eta to be somewhere this week.

PR #381, opened 2018-10-03
FB custom scope + upgrade api version

doron2402:
- Upgrade FB version from `2.9` to `3.1`
- Add the ability to update Facebook login scope


doron2402:
@AdriVanHoudt anything I should add?

AdriVanHoudt:
@doron2402 looks good, is there a breaking api change between these versions?

doron2402:
thanks @AdriVanHoudt did some testing and double checked Facebook api docs seems fine:
https://developers.facebook.com/docs/graph-api/reference/v3.1/user

doron2402:
oops I mean to comment not to close it.

doron2402:
@AdriVanHoudt can we merge this?

AdriVanHoudt:
I can only see https://developers.facebook.com/docs/graph-api/changelog/version3.0#login which I don't think is breaking in our case.

PR #396, opened 2019-02-11
Allow isSameSite customization

### LINKED ISSUE #395: Safari isSameSite
I'm having redirect loops upon trying to login with Okta in Safari. I noticed that if I manually change `isSameSite` to `'Lax'` or `false` in the `cookieOptions` for bell it fixes the issue. 

Bell: 9.4.0
Hapi: 18.1.0
Hapi auth cookie: 9.1.0
Safari: 12.0.2

flipxfx:
Fixes #395 by allowing customization of the `isSameSite` param.

### DIFF
--- a/API.md
+++ b/API.md
@@ -147,6 +147,7 @@ Each strategy accepts the following optional settings:
 - `cookie` - the name of the cookie used to manage the temporary state. Defaults to
   `'bell-provider'` where 'provider' is the provider name (or `'custom'` for custom providers). For
   example, the Twitter cookie name defaults to `'bell-twitter'`.
+- `isSameSite` - sets the cookie same site option. Defaults to `Strict`.
 - `isSecure` - sets the cookie secure flag. Defaults to `true`.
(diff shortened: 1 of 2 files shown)


AdriVanHoudt:
Looks good, thanks!

Is it ok if we only allow `false`, `Lax` or `Strict` since those are the only ones allowed by hapi?

flipxfx:
@AdriVanHoudt Yeah I had thought about that, but I wasn't sure what the standard was for that. Since hapi will catch that, I figured it was less maintenance if we make the rule more lax in bell.

flipxfx:
Is there a timeline of when the next release will be?

AdriVanHoudt:
I got caught up in looking into supporting hapi 18 but if you want I can do a minor release to get this out before that.

flipxfx:
That would be sweeeeeet.

AdriVanHoudt:
Done 🚀

PR #397, opened 2019-02-27
Add support for hapi 18

### LINKED ISSUE #394: Google auth redirect loop with hapi v18
Upgrading from v17.8.1 -> v18.0.0 seems to cause a redirect loop. Following the same exact configuration pattern as https://github.com/hapijs/bell/blob/master/examples/google.js

My router handler is never actually called, instead the route is redirected to the google SSO login page, and when the correct google account is selected, it redirects back, but the handler is never called. Instead it redirects back to Google. Downgrading to 17.8.1 it works fine.

rluba:
This PR includes two fixes related to [hapi 18](https://github.com/hapijs/hapi/issues/3871):

1. Many tests broke because hapi now strips the default port from `request.info.host` (a side effect of using the WHATWG URL API) – causing lots of assertions to fail. I decided to use a non-default port in the tests instead of removing the default port from the assertions. This verifies that port info is still propagated correctly.
2. `request.url.query` is no longer available. I’ve changed it to `request.query` which works for hapi 18 and older versions. (I’ve tested it with hapi 17 and 18). This fixes #394.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -353,7 +353,7 @@ internals.refreshRedirect = function (request, name, protocol, settings, credent
         return h.unauthenticated(Boom.internal('Missing ' + name + ' request token cookie'), { credentials });
     }
 
-    const refreshQuery = Object.assign({}, request.url.query, { refresh: 1 });
+    const refreshQuery = Object.assign({}, request.query, { refresh: 1 });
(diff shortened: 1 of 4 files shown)


AdriVanHoudt:
This is amazing, thank you!

rluba:
@AdriVanHoudt You’re welcome. Thank you for maintaining this project. 🙇‍♂️

AdriVanHoudt:
No problem! It is PR's like this that make it possible :O

Issue #376, opened 2018-07-24
Failed obtaining office365 user profile. v8.8.0


freakinside:
Im using bell version 8.8.0

I have github integration working but i have problems with office365. After i select a user it seems there is a problem obtaining the profile.

I checked the code on the latest version and seems none of the specific urls for office365 changed, so i assume version 8.8.0 should still work as other providers do.

{ Error: Failed obtaining office365 user profile
    at Wreck.(anonymous function) (/var/app/current/node_modules/bell/lib/oauth.js:298:43)
    at read (/var/app/current/node_modules/wreck/lib/index.js:523:20)
    at finish (/var/app/current/node_modules/wreck/lib/index.js:344:20)
    at wrapped (/var/app/current/node_modules/hoek/lib/index.js:879:20)
    at onReaderFinish (/var/app/current/node_modules/wreck/lib/index.js:415:16)
    at g (events.js:292:16)
    at emitNone (events.js:91:20)
    at emit (events.js:185:7)
    at finishMaybe (_stream_writable.js:512:14)
    at endWritable (_stream_writable.js:522:3)
    at Writable.end (_stream_writable.js:487:5)
    at IncomingMessage.onend (_stream_readable.js:511:10)
    at IncomingMessage.g (events.js:292:16)
    at emitNone (events.js:91:20)
    at IncomingMessage.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:974:12)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
  data: <Buffer >,
  isBoom: true,
  isServer: true,
  output: 
   { statusCode: 500,
     payload: 
      { statusCode: 500,
        error: 'Internal Server Error',
        message: 'An internal server error occurred' },
     headers: {} },
  reformat: [Function] }

freakinside:
In /var/app/current/node_modules/bell/lib/oauth.js 

Wreck.post(settings.provider.token, requestOptions, (err, tokenRes, payload) => {

It assumes that payload object has access_token and refresh_token, but i have id_token and refresh_token instead when using office365 provider. Not sure why i dont get an access_token as response code is 200.

freakinside:
I already get the access_token. The problem is that there the default scopes are note enough to get an access_token. I solved it adding 'User.Read'.

```
    server.auth.strategy('office365', 'bell', {
        provider: 'office365',
        password: Config.auth.password,
[...]

PR #399, opened 2019-03-04
 Fix auth info for office365 

freakinside:
In the outlook api documentation now points to graph api to retrieve people information.

### DIFF
--- a/lib/providers/office365.js
+++ b/lib/providers/office365.js
@@ -18,12 +18,12 @@ exports = module.exports = function (options) {
         scope: ['openid', 'offline_access', 'profile'],
         profile: async function (credentials, params, get) {
 
-            const profile = await get('https://outlook.office.com/api/v2.0/me');
+            const profile = await get('https://graph.microsoft.com/v1.0/me');
 
             credentials.profile = {
-                id: profile.Id,
-                displayName: profile.DisplayName,
-                email: profile.EmailAddress,
+                id: profile.id,
+                displayName: profile.displayName,
+                email: profile.mail,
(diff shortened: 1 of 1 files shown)


AdriVanHoudt:
I understand this is a thing Bell needs, have you read the tail of https://github.com/hapijs/bell/issues/376 and what do you think of that? And the solution I proposed in https://github.com/hapijs/bell/issues/376#issuecomment-422704152?

freakinside:
Sorry. I didn't have the time to return to this issue before.

In #376 you propose a posibility to fork the provider to have a "microsoft" one. The fix i made to "office365" is what i think you proposed for that one.

We have to keep in mind that "https://outlook.office.com/api/v2.0/me" is not working atm if you only want the profile as it crashes due to the missing token. In the past it worked with a weird combination of scopes but it seems not to work anymore (not sure what changed).

Maybe the solution is just to rename the provider and document it, so people that had used the "office365" provider in the past find the reason and double check what scopes may need now.

AdriVanHoudt:
Thanks for the reply!
No problem on the reply time, as you might have noticed I also don't have a lot of time :D
I'm working on a solution to clean up all the Microsoft oauth stuff. The main "problem" I have with the change in this PR is that it would be breaking. While adding a new one like https://github.com/hapijs/bell/pull/404 would not be.
I'm thinking about adding a new one and recommend that one and document it everywhere that the others are deprecated and they should probably use the new one. How does that sound?

freakinside:
Probably thats the best way to do it. That way no one would be surprised.

Issue #401, opened 2019-03-14
Passing Scope


imcodingideas:
Inside `/lib/providers/slack.js#L24` I need to be able to pass a different scope to profile.

        profile: async function (credentials, params, get) {

            credentials.profile = {
                access_token: params.access_token,
                scope: params.scope
            };
        }

Any user who is not an admin seems to be getting:         `scope: ['identify'],                                    // Minimum scope required to identify the user` and admins are getting the proper scope. 

This means I need to pass it.. in my case: `scope: ['identify', 'groups:read', 'channels:read', 'im:read', 'mpim:read', 'users:read', 'users:read.email', 'users.profile:read']`

https://github.com/hapijs/bell/blob/master/examples/slack.js

How do I pass the scope to the function?

imcodingideas:
@AdriVanHoudt is there any thing I can to help?

AdriVanHoudt:
Where you pass your config to `server.auth.strategy` you can set the `scope` to what you want. See https://github.com/hapijs/bell/blob/master/API.md#options

hueniverse:
If someone wants to improve the docs, I'll merge.

PR #402, opened 2019-03-15
Updates slack example

imcodingideas:
I propose this documentation change as it took hours to figure out and hopefully this will save someone else hours.

I was having issues with the redirect call back on my local machine. For some reason adding: isSameSite: 'Lax', fixed it. I have no idea why but for me it was redirecting from slack to my local endpoint. I think the example could have it like this or committed out. 

https://stackoverflow.com/questions/39748688/hapi-js-bell-auth-cookie-redirect-loop

The second scope array. This closes: https://github.com/hapijs/bell/issues/401 but basically, non admin users were getting the default scope: `scope: ['identify'], // Minimum scope required to identify the user` so with this I think it will show users how to let their non admin users pass custom scopes.

### DIFF
--- a/examples/slack.js
+++ b/examples/slack.js
@@ -16,7 +16,18 @@ internals.start = async function () {
         password: 'cookie_encryption_password_secure',
         isSecure: false,
         clientId: '',
-        clientSecret: ''
+        clientSecret: '',
+        isSameSite: 'Lax', // fixes issue with Auth Redirect Loop
+        scope: [
+          'identify',
+          'groups:read',
+          'channels:read',
+          'im:read',
+          'mpim:read',
+          'users:read',
+          'users:read.email',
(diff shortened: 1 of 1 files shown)


AdriVanHoudt:
Hi!
Sorry you had to struggle with using Bell :/
If these changes would have helped you I would be glad to add them.
Would you be ok with putting them in quoted? I would like people to try the default provider and then if it doesn't work try these other options. If everyone has issues with the default we might need to update that instead of the docs. Does that make sense?

imcodingideas:
Yea, I will update my PR with that!

hueniverse:
To be honest, I don't understand the comments. I get that these changes solved your problem, but I can't see how they will help someone else. Instead, I've added a detailed notes section to the API that I think would have helped you.

PR #409, opened 2019-05-02
RFC 7636: Proof Key for Code Exchange

adaxi:
PKCE ([RFC 7636](https://tools.ietf.org/html/rfc7636)) is a technique to secure public clients that don't use a client secret. 

It is primarily used by native and mobile apps, but the technique can be applied to any public client as well.

Although `hapi` and `bell` are usually used in places where a secret can be kept safely, someone might be using it as a public client. 

My motivation for this pull request is not because my application is public, but because I am using a provider which requires PKCE.

More information:
https://oauth.net/2/pkce/
https://www.oauth.com/oauth2-servers/pkce/
https://www.oauth.com/oauth2-servers/pkce/authorization-request/

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -49,6 +49,7 @@ internals.schema = Joi.object({
         headers: Joi.object(),
         profile: Joi.func(),
         profileMethod: Joi.string().valid('get', 'post').default('get'),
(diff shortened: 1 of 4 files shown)


hueniverse:
Interesting. I'll try to get to it next couple of weeks.

adaxi:
bump.

PR #445, opened 2019-12-11
Use actual package name for plugin name

sholladay:
This PR fixes a problem I am having during my upgrade to hapi 18 and the new `@hapi` scoped package names.

My [hapi-doorkeeper](https://github.com/sholladay/hapi-doorkeeper) plugin [depends on](https://github.com/sholladay/hapi-doorkeeper/blob/master/index.js#L123-L126) `@hapi/cookie` and `@hapi/bell`. However, the `@hapi/bell` dependency was failing with this error when starting the server...

```js
Error: Plugin hapi-doorkeeper missing dependency @hapi/bell
```

It took a while to figure out that it's because `@hapi/bell` has a hardcoded plugin name that is different from the package name, which is inconsistent with `@hapi/cookie` and others. I've fixed this to be consistent with other plugins, in that the `@hapi/` prefix is part of the plugin name.

This is technically a breaking change. However, I think it may have been unintentionally omitted from bell v11, so in that sense it's a just a patch.

### DIFF
--- a/examples/okta.js
+++ b/examples/okta.js
@@ -11,7 +11,7 @@ const internals = {};
 internals.start = async function () {
 
     const server = Hapi.server({ port: 8000 });
-    await server.register([require('hapi-auth-cookie'), Bell]);
+    await server.register([require('@hapi/cookie'), Bell]);
 
     server.auth.strategy('session', 'cookie', {
         password: 'secret_cookie_encryption_password',      // Use something more secure in production
--- a/lib/index.js
+++ b/lib/index.js
@@ -23,7 +23,6 @@ exports.oauth = OAuth;
 // Plugin
 
 exports.plugin = {
-    name: 'bell',
     pkg: require('../package.json'),
     requirements: {
         hapi: '>=18.3.0'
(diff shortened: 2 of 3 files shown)


sholladay:
Looks like this is also the case in [vision](https://github.com/hapijs/vision/blob/53d4414c94e13060a573f190944d416bac28f52d/lib/index.js#L16), [inert](https://github.com/hapijs/inert/blob/932202395f58e3da76e397d6151049a0148d02fe/lib/index.js#L35), and [h2o2](https://github.com/hapijs/h2o2/blob/891cd9c8ba1836a6fdfc7eb5383462842fd0188d/lib/index.js#L65), probably others. Some of these have a code comment that specifically says "override package name". So I guess it's intentional and `@hapi/cookie` was accidentally published _with_ the `@hapi` scope included in the plugin name.

I definitely see the benefit of `server.plugins.bell` over `server.plugins['@hapi/bell']`, but whatever the choice is, it should be consistent.

On Slack, Eran proposed an alternative solution:
> I would actually special case `@hapi` in the name comparison within hapi core...

PR #450, opened 2020-01-13
Support Okta custom authorization server

timcosta:
[Okta custom authorization servers](https://support.okta.com/help/s/article/Difference-Between-Okta-as-An-Authorization-Server-vs-Custom-Authorization-Server) require the addition of an alphanumeric ID after the `/oauth2` URL segment which cannot be done using just the base `uri` option. I opted to implemented this change this way as to not require a major version bump, even though it might be cleaner to just drop the `/oauth2` part of the URL that Bell adds and force people to move it into their `uri` option.

### DIFF
--- a/Providers.md
+++ b/Providers.md
@@ -842,6 +842,7 @@ credentials.profile = {
 - `scope`: Defaults to `['openid', 'email', 'offline_access']`
 - `config`:
   - `uri`: Point to your Okta enterprise uri.  Intentionally no default as Okta is organization specific.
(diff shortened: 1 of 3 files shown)


hueniverse:
I think we should just add a new okta provider with the new feature and when we do another major, make that the default.

timcosta:
I'm happy to wait on this until the next major release. I'm not actively using this and don't need it, I was just building a PoC against Okta and realized that this was missing so I figured I would PR it.

kPOWz:
would love to see in 12.1.0

PR #452, opened 2020-01-22
OAuth1: Added support for PLAINTEXT signature

das-peter:
Would be great if the PLAINTEXT signature as defined in https://tools.ietf.org/html/rfc5849#section-3.4.4 is supported too.
Seems pretty straight forward.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -49,7 +49,7 @@ internals.provider = Joi.object({
         is: 'oauth',
         then: Joi.object({
             temporary: Joi.string().required(),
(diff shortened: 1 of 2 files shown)


hueniverse:
What for?!

das-peter:
>What for?!

To proceed to feature completeness -  because there are APIs like https://developer.trademe.co.nz/api-overview/authentication/example-plaintext-workflow/ that straight forward recommend to use PLAINTEXT(?!?).
I personally don't need it because I added it on the go while checking out the module & the above mentioned API.
However, if not shot down beforehand I'm happy to add the required tests.

hueniverse:
I don't want to support an insecure option. No one should be using it.

PR #456, opened 2020-02-18
Add authentication information documentation

jonathansamines:
Adding missing documentation for `request.auth` schema for all supported protocols and authentication configurations.

**Note:** Wasn't sure if  `request.auth.credentials.profile` required documentation as it was already documented on each provider and the property name is not mandatory (a custom provider might use a different name)


hueniverse:
Thanks!

PR #463, opened 2020-06-08
Support functions for `provider.auth` and `provider.token`

### LINKED ISSUE #461: Support functions for `provider.auth` and `provider.token`
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): Community
* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: latest
* *module version*: latest
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): `@hapi/hapi@latest`
* *any other relevant information*: n/a

#### What problem are you trying to solve?

I'm attempting to use this module in a Shopify app that integrates with [Shopify's OAuth2 authentication workflow](https://shopify.dev/tutorials/authenticate-with-oauth).

Shopify works slightly differently from the current providers in that the authorization step is _per store_, such that the `provider.auth` and `provider.token` options differ depending on the store that initiated the OAuth flow (AKA the store that installed the "app").

In other words, in order for "Store A" authenticate it must redirect to `https://store-a.myshopify.com/admin/oauth/authorize`, while a "Store B" must redirect to `https://store-b.myshopify.com/admin/oauth/authorize`.

#### Do you have a new or modified API suggestion to solve the problem?

As far as I can tell, [`@shopify/koa-shopify-auth`](https://github.com/Shopify/quilt/tree/master/packages/koa-shopify-auth) is able to determine which store to redirect to through a query parameter in the request. I think a similar approach could be taken here.

Would you support expanding the `provider` schema so that [`auth` and `token` options](https://github.com/hapijs/bell/blob/4ef909088296538eff5de0e8f6a99cb10e8dc431/lib/index.js#L42-L43) can alternatively accept functions with access to `request`, similar to the `providerParams` or `scope` options (#233)?

Thanks!

charlesbjohnson:
Resolves #461

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -39,8 +39,8 @@ exports.plugin = {
 internals.provider = Joi.object({
     name: Joi.string().optional().default('custom'),
     protocol: ['oauth', 'oauth2'],
-    auth: Joi.string().required(),
-    token: Joi.string().required(),
+    auth: Joi.alternatives(Joi.string(), Joi.func()).required(),
+    token: Joi.alternatives(Joi.string(), Joi.func()).required(),
     headers: Joi.object(),
     profile: Joi.func(),
     profileMethod: Joi.valid('get', 'post').default('get')
@@ -48,7 +48,7 @@ internals.provider = Joi.object({
     .when('.protocol', {
         is: 'oauth',
         then: Joi.object({
-            temporary: Joi.string().required(),
+            temporary: Joi.alternatives(Joi.string(), Joi.func()).required(),
             signatureMethod: Joi.valid('HMAC-SHA1', 'RSA-SHA1').default('HMAC-SHA1')
         }),
         otherwise: Joi.object({
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -23,6 +23,11 @@ exports.v1 = function (settings) {
 
     return async function (request, h) {
 
+        client.settings.temporary = client.settings.temporary ||
+          internals.Client.baseUri(internals.resolveProviderEndpoint(request, settings.provider.temporary));
+        client.settings.token = client.settings.token ||
+          internals.Client.baseUri(internals.resolveProviderEndpoint(request, settings.provider.token));
+
         const cookie = settings.cookie;
         const name = settings.name;
         const protocol = internals.getProtocol(request, settings);
@@ -56,10 +61,9 @@ exports.v1 = function (settings) {
 
             // Obtain temporary OAuth credentials
 
-            const temporaryEndpoint = internals.resolveProviderEndpoint(request, settings.provider.temporary);
             const oauth_callback = internals.location(request, protocol, settings.location);
             try {
-                var { payload: temp } = await client.temporary(temporaryEndpoint, oauth_callback);
+                var { payload: temp } = await client.temporary(oauth_callback);
             }
             catch (err) {
                 return h.unauthenticated(err, { credentials });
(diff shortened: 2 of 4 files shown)


lib/index.js
```diff
@@ -39,16 +39,16 @@ exports.plugin = {
...
+    token: Joi.alternatives(Joi.string(), Joi.func()).required(),
     headers: Joi.object(),
     profile: Joi.func(),
     profileMethod: Joi.valid('get', 'post').default('get')
 })
     .when('.protocol', {
         is: 'oauth',
         then: Joi.object({
-            temporary: Joi.string().required(),
+            temporary: Joi.alternatives(Joi.string(), Joi.func()).required(),
```
charlesbjohnson, line 51: My interest is only in `auth` and `token`, but I thought it might be odd if I had left this one out.

test/oauth.js
```diff
@@ -2154,6 +2235,12 @@ describe('Bell', () => {
                 expect(OAuth.Client.baseUri('http://example.com:8080/x')).to.equal('http://example.com:8080/x');
                 expect(OAuth.Client.baseUri('https://example.com:8080/x')).to.equal('https://example.com:8080/x');
             });
+
+            it('passes through without port', () => {
+
+                expect(OAuth.Client.baseUri('http://example.com/x')).to.equal('http://example.com/x');
+                expect(OAuth.Client.baseUri('https://example.com/x')).to.equal('https://example.com/x');
+            });
```
charlesbjohnson, line 2243: This is unrelated, but after my changes we lost coverage on this case. This adds it back.

lib/oauth.js
```diff
@@ -387,28 +393,26 @@ exports.Client = internals.Client = function (options) {
...
-        token: internals.Client.baseUri(options.provider.token),
         clientId: options.clientId,
         clientSecret: options.provider.signatureMethod === 'RSA-SHA1' ? options.clientSecret : internals.encode(options.clientSecret || '') + '&',
         signatureMethod: options.provider.signatureMethod
     };
 };
 
 
-internals.Client.prototype.temporary = function (oauth_callback) {
+internals.Client.prototype.temporary = function (uri, oauth_callback) {
```
hueniverse, line 403: This is a breaking change. The `Client` is a public interface. Can't change it.

hueniverse (changes requested):

lib/oauth.js
```diff
@@ -387,28 +393,26 @@ exports.Client = internals.Client = function (options) {
...
         oauth_callback
     };
 
-    return this._request('post', this.settings.temporary, null, oauth, { desc: 'temporary credentials' });
+    return this._request('post', internals.Client.baseUri(uri), null, oauth, { desc: 'temporary credentials' });
 };
 
 
-internals.Client.prototype.token = function (oauthToken, oauthVerifier, tokenSecret) {
+internals.Client.prototype.token = function (uri, oauthToken, oauthVerifier, tokenSecret) {
```
hueniverse: Same on every `Client` endpoint... I thought that was implicit... :-)

hueniverse (changes requested):

lib/oauth.js
charlesbjohnson: I reverted the changes to these method signatures but kept their ability to resolve endpoints from a function, if a function is provided. It's a bit ugly since it potentially requires setting `client.settings.token` and `client.settings.temporary` in two passes, but I think it only applies to those who choose to upgrade to using a function instead of a string.

I didn't catch that `Client` was exported before (d'oh 😅). I think this should resolve the breaking change, right? Is there something else that I'm missing?

hueniverse:
Thanks for the changes. I hope to get to it in the next few days. If not, please ping me.

imcodingideas (approved):

Marsup:
Thanks for your contribution, considering this PR was 6 years old and I had a few more comments to make, I had little hope that you'd update it, so I went with mine, hope you don't mind.

PR #466, opened 2020-08-03
upgrade bell to use just regular joi package and update teamwork as well

lloydbenson:


Marsup:
Shouldn't you be using hapi/validate ?

lloydbenson:
can't because hapi/validate does not have the Joi.string().uri() function anymore that these rely on unfortunately.

cjihrig:
Also worth noting that the "no external code" policy is only for core. Plugins are a bit more relaxed, and many of them already have external dependencies.

PR #468, opened 2020-08-13
Node 14

Nargonath:
I've updated the Travis configuration to use travis templates.

Do we want to switch `bell` to `@hapi/validate`? I included a commit for it but it breaks as `bell` is using `.string().uri()` which is not part of `@hapi/validate`.


Nargonath:
Bell will continue to use `joi`. My bad, forgot that @lloydbenson already had the same issue. I removed my commit from the PR.

PR #469, opened 2020-08-20
[Bugfix] Okta example code

doron2402:
The example throws an error due to change in @hapi/cookie [API ](https://github.com/hapijs/cookie/blob/master/lib/index.js#L32-L54)


lloydbenson:
thx for the fix!

PR #476, opened 2021-01-22
Add support for providing clientSecret as a function

### LINKED ISSUE #462: Sign in with apple
#### Support plan

* *which support plan is this issue covered by?* (e.g. Community, Core, Plus, or Enterprise): Community
* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: latest
* *module version*: latest
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...):
* *any other relevant information*: n/a

#### What problem are you trying to solve?

I'm setting up "Sign in with Apple" with bell and hapi.

I have made a custom provider and gotten everything to work, but there are a few problems.

The biggest problem is apple uses a JWT as clientSecret, which have an expire date.

My apple strategy looks like this right now.
```js
    server.auth.strategy('apple', 'bell', {
      clientId: oAuthClients.apple.clientId,
      clientSecret: getSecretKey(),
      provider: apple,
      providerParams: { response_mode: 'query' },
    })
```

The `getSecretKey()` generates a JWT that expires in six months and then it will stop working (if it doesn't get restarted)

Next problem; apple doesn't have a getProfile-url, instead they return a JWT with the user email when authenticating.

My custom provider:
```js
const apple = {
  auth: 'https://appleid.apple.com/auth/authorize',
  token: 'https://appleid.apple.com/auth/token',
  name: 'apple',
  protocol: 'oauth2',
  useParamsAuth: true,
  profile: async (credentials, params) => {
    const resp = jwt.verify(params.id_token, getSecretKey())
    const { sub, email } = resp
    credentials.profile = { id: sub, email }
    return credentials
  },
}
```
As you can see in the profile function I use a `jwt.verify` and the clientSecret to get the token payload.

#### Do you have a new or modified API suggestion to solve the problem?

Making the `clientSecret` take a function could fix the token expire issue.

Getting the profile issue could be solved-ish by adding the clientSecret to the profile params, but it would still require a jwt library.

If adding a jwt library is a possibility I think it would make sense to use that to create the clientSecret as well.

Another possibility could be to add a verify/decode-token function to the config object and use that to get the user profile (dependency injection?).
Then it wouldn't be necessary to add a jwt library to bell.

The getSecretKey for reference
```
const getSecretKey = () => {
    const claims = {
      iss: teamId,
      iat: Math.floor(Date.now() / 1000),
      exp: Math.floor(Date.now() / 1000) + 86400 * 180,
      aud: 'https://appleid.apple.com',
      sub: appId,
    }

    const token = sign(claims, privateKey, {
      algorithm: 'ES256',
      keyid: keyId,
    })

	return token
}
```

I will gladly make a pull request for an apple provider, if you want that.

Ginden:
Partially resolves #462

### DIFF
--- a/API.md
+++ b/API.md
@@ -226,6 +226,8 @@ The `server.auth.strategy()` method requires the following strategy options:
   object will be merged with the Wreck request object used to call the token endpoint. Such an
   object can contain custom HTTP headers or TLS options (e.g.
   `{ agent: new Https.Agent({ cert: myClientCert, key: myClientKey}) }`).
+  To allow dynamically updating secret, this option can be passed as a *function* returning string
+  to be used as `clientSecret`.
 - `forceHttps` - A boolean indicating whether or not you want the redirect_uri to be forced to
   https. Useful if your hapi application runs as http, but is accessed through https.
 - `location` - Set the base redirect_uri manually if it cannot be inferred properly from server
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -263,6 +263,9 @@ exports.v2 = function (settings) {
             if (typeof settings.clientSecret === 'string') {
                 query.client_secret = settings.clientSecret;
             }
+            else if (typeof settings.clientSecret === 'function') {
+                query.client_secret = settings.clientSecret();
+            }
         }
 
         const requestOptions = {
--- a/lib/index.js
+++ b/lib/index.js
@@ -72,6 +72,7 @@ internals.schema = Joi.object({
 
     clientSecret: Joi.alternatives()
         .try(Joi.string().allow(''))
+        .try(Joi.function())
         .conditional('provider.protocol', {
             not: 'oauth',
             then: Joi.object()
--- a/test/oauth.js
+++ b/test/oauth.js
@@ -928,6 +928,48 @@ describe('Bell', () => {
             expect(res.headers.location).to.contain(mock.uri + '/auth?special=true&client_id=test&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A8080%2Flogin&state=');
         });
 
+        it('uses clientSecret() function', async (flags) => {
+
+            const mock = await Mock.v2(flags);
+            const server = Hapi.server({ host: 'localhost', port: 8080 });
+            await server.register(Bell);
+
+            let callCount = 0;
+            const secretGetter = () => {
+
+                callCount += 1;
+                return 'secret';
+            };
+
+            server.auth.strategy('custom', 'bell', {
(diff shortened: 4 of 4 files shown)


API.md
```diff
@@ -226,6 +226,8 @@ The `server.auth.strategy()` method requires the following strategy options:
   object will be merged with the Wreck request object used to call the token endpoint. Such an
   object can contain custom HTTP headers or TLS options (e.g.
   `{ agent: new Https.Agent({ cert: myClientCert, key: myClientKey}) }`).
+  To allow dynamically updating secret, this option can be passed as a *function* returning string
```
Nargonath, line 229: We should mention that the fonction is called synchronously.

lib/oauth.js
```diff
@@ -263,6 +263,9 @@ exports.v2 = function (settings) {
             if (typeof settings.clientSecret === 'string') {
                 query.client_secret = settings.clientSecret;
             }
+            else if (typeof settings.clientSecret === 'function') {
+                query.client_secret = settings.clientSecret();
```
Nargonath, line 267: Is there any parameters that could be interesting to pass on here?

Ginden: `settings`?

API.md
Ginden: I think _function returning string_ is equivalent to `() => string` in TypeScript, not `() => string | Promise<string>`. Though, we can just add `await` in `lib/oauth.js:267`

Nargonath: I agree but we didn't express any type definition here. This was just to prevent people from using async function and then wonder why it does not work the way they intended to. I don't expect every user to look at the code. I'm in favor of leaving sync for now especially you don't seem to need it. We can always reassess later if someone needs it.

lib/oauth.js
Nargonath: I'm not much of a bell user myself that's just a question that crossed my mind. I just wondered if that's something you considered on your end. If you don't need it let's keep it that way then.

Ginden:
This blocks our application development. @Nargonath

Ginden:
Any update on this? @Nargonath

Nargonath:
Sorry for the lack of udpate on this @Ginden. I was waiting for others to chime in. I'll try to get some feedbacks and we'll see. 😃

devinivy (approved):

Nargonath (approved):

Nargonath:
It has been deployed as `v12.2.0`. Thanks @Ginden !

https://github.com/hapijs/bell/releases/tag/v12.2.0

PR #477, opened 2021-02-02
Added state before redirecting from 'refreshRedirect' function

Nithin-Philips:


Nargonath:
@Nithin-Philips would you mind detailing the reasons behind your MR, please? 🙂 Thank you.

PR #481, opened 2021-08-20
Support OAuth1 client wreck options

hueniverse:
Adds support for passing Wreck options to allow setting agent and other overrides.

### DIFF
--- a/lib/oauth.js
+++ b/lib/oauth.js
@@ -396,6 +396,8 @@ exports.Client = internals.Client = function (options) {
         clientSecret: options.provider.signatureMethod === 'RSA-SHA1' ? options.clientSecret : internals.encode(options.clientSecret || '') + '&',
         signatureMethod: options.provider.signatureMethod
     };
+
+    this._wreckOptions = { ...options.wreck };
 };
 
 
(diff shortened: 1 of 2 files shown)


lib/oauth.js
```diff
@@ -451,6 +453,7 @@ internals.Client.prototype._request = async function (method, uri, params, oauth
     // Calculate OAuth header
 
     const requestOptions = {
+        ...this._wreckOptions,
         headers: {
             Authorization: internals.Client.header(oauth)
         }
```
Nargonath, line 459: Wouldn't you want also to be able to pass extra headers other than `Authorization`?

I was thinking of something like that to accomodate for it:

```js
const requestOptions = {
        ...this._wreckOptions,
        headers: {
            ...this._wreckOptions.headers,
            Authorization: internals.Client.header(oauth)
        }
};
```

Nargonath (commented):
Shouldn't we document that behavior in the options section of the API.md? Apart from that and my little comment, LGTM.

lib/oauth.js
hueniverse: Not a use case I've seen. I don't mind it :-)

hueniverse:
The client is not documents AFAICT which is why I didn't add that new option.

PR #484, opened 2021-12-12
add dispatcher and use shared config

lloydbenson:


Nargonath (commented):
Seems we have some failing tests, not sure they're flaky or something that needs to be fixed. I haven't looked deeply yet.

PR #485, opened 2022-01-20
remove the bit about google plus

capaj:
I don't think this still exists in the google cloud console


Nargonath:
Google+ API is still in the Google Cloud Console:

image

From what I understood you need to active the API as a whole but perhaps I'm mistaken here.

mkg20001:
google+ is deprecated and currents (enterprise google+) is being [sunset](https://chromeunboxed.com/google-currents-shut-down-announcement) in favor of google chat spaces (not the other spaces...) so I guess it's dead. The docs links leads to a 404, I think it's going to be removed as soon as currents is dead.

PR #488, opened 2022-10-10
Support hapi v21 and node v18, drop hapi v19 and node v12, test ESM support

devinivy:
- Support and test on hapi v21, drop support for hapi v19 and below.
 - Support and test on node v18, drop support for node v12 and below.
 - Test ESM support.
 - Utilize some node v14+ new syntax.

If this looks good, this will go out as bell v13.


Nargonath (approved):
Thanks for the work Devin.

PR #494, opened 2024-03-30
Pass user credentials

harryadel:
When you provide `scope` property for apple authentication. Apple responds with the user data in the authentication request and not the authorization request. So this PR aims to provide these values back so it can be used in the `profile` property.

https://dev.to/harryadel/apple-sign-in-using-hapijs-32ik


Marsup:
Better late than never I guess, I [implemented this](https://github.com/hapijs/bell/pull/502) by exposing the request to the profile hook.

PR #497, opened 2025-03-11
Add tokenParams option

kanongil:
This mirrors the `providerParams` and `profileParams` options, to allow custom parameters to be supplied to the `token` endpoint request.

This is needed by Auth0, which cannot return a JWT token without the `endpoint` parameter added to both the `/authorize` and `/token` request. More details here: https://auth0.com/docs/secure/tokens/access-tokens/get-access-tokens#control-access-token-audience.

Note that the existing `providerParams` option, seems to be confusingly named. I guess `authParams` would be better, but renaming it requires a breaking change.


kanongil:
The test failures does not seem related to this PR.

Marsup (approved):
LGTM

